PR #20, opened 2023-08-14
Export base-button

jd-carroll:
There are many places where it would be useful to have the BaseButton when developing custom button components.


jd-carroll:
Actually, after doing a little more investigation, maybe this isn't necessary...

Leaving open for now

PR #23, opened 2023-08-15
Fix grid prop defs (docs)

jd-carroll:
Grid already supports responsive rows / cols (prop defs were inaccurate)

### DIFF
--- a/packages/radix-ui-themes/src/components/grid.props.ts
+++ b/packages/radix-ui-themes/src/components/grid.props.ts
@@ -1,4 +1,4 @@
-import type { PropDef } from '../helpers';
+import type { PropDef, ResponsivePropDef } from '../helpers';
 
 const displayValues = ['none', 'inline-grid', 'grid'] as const;
 const flowValues = ['row', 'column', 'dense', 'row-dense', 'column-dense'] as const;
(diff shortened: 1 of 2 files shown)


vladmoroz:
I am not seeing any issues with types when using Grid. What issue does this solve exactly?

jd-carroll:
Check the docs on the website, neither does or columns is listed as
responsive

vladmoroz:
Oh, documentation is separate in https://github.com/radix-ui/website. Thanks for the note, we'll update it

jd-carroll:
@vladmoroz I believe the way the website works is that it picks up the `propDefs` of the individual component and then renders that in a table.
https://github.com/radix-ui/website/blob/ee246a5e0451744e2f127a006726c70e4727be4b/components/ThemesPropsTable.tsx#L37

So doesn't it need to be changed here first, and then have the updated module installed for the website to be picked up?

vladmoroz:
It does, but the docs use the value of the prop defs (which already have an explicit `responsive: true` value for columns and rows props) and not their type signature, which is what this PR adjusts—this would have no effect on the docs anyway.

Since this package itself defines these props correctly, it's something we need to fix in the website code

Issue #22, opened 2023-08-15
The dark theme has not worked properly.


zmzimpl:
It seems that the issue arises after setting appearance="dark" without properly setting the HTML attributes class="light-theme" and style="color-scheme: light;".

![image](https://github.com/radix-ui/themes/assets/27490895/42eaeaef-8f72-463a-ab64-68820247be2c)

zmzimpl:
`@radix-ui/themes`  version: 1.0.0

zmzimpl:
I cloned the source code attempting to fix this issue, but encountered an error when running it:
```
D:\dev\github\themes>npm run dev  

> dev
> turbo run dev --no-cache --continue

• Running dev
• Remote caching disabled
root task dev (turbo run dev --no-cache --continue) looks like it invokes turbo and might cause a loop

No tasks were executed as part of this run.

 Tasks:    0 successful, 0 total
Cached:    0 cached, 0 total
  Time:    171ms

 ERROR  run failed: command  exited (1)
```
I'm wondering what the issue is, as I couldn't find an answer on Google. Perhaps you might know?

jd-carroll:
I can confirm this behavior, here is a codesandbox which demonstrates it:
https://codesandbox.io/s/intelligent-framework-2xqjp8?file=/src/App.tsx

The issue is here:
https://github.com/radix-ui/themes/blob/39f5a438db0d8496134d4bc35d71899fee1b2562/packages/radix-ui-themes/src/theme.tsx#L152-L159

When the root theme element is rendered, the `useContext` call will still return undefined. So if you look at the example, it actually appears to be using some gray palette and not an actual dark theme. 

If you move the `appearance="dark"` to the second `Theme` component, it works as expected.

Not sure what the best fix would be, but a fix would be changing the [appearance prop](https://github.com/radix-ui/themes/blob/39f5a438db0d8496134d4bc35d71899fee1b2562/packages/radix-ui-themes/src/theme.tsx#L159) to something like:
```typescript
  appearance = props.appearance ?? context?.appearance ?? themePropDefs.appearance.default,
```

This way, the Theme component always prefers component props and then falls back to context and default values.

Note: A temporary workaround would be to do something like:
```jsx
<Theme appearance="dark">
  <Theme>
    <MyApp />
  </Theme>
</Theme>
```

(Looks ridiculous, but ensures the second theme receives the desired props)

jd-carroll:
[...]

PR #32, opened 2023-08-18
Extended theme

jd-carroll:
This PR proposes a number of changes to the core Theme element. These changes were necessary as part of the implementation of the [MUI Grid v2 (fluid grid)](https://mui.com/material-ui/react-grid2).

Specifically, this PR addresses these issues:
- Need for programmatic breakpoint access (achieved through `breakpoints-context.tsx`)
- Need for open-ended spacing based on a common factor (achieved with a new prop on Theme and `spacing.ts`)

Additionally, this PR proposes a fix for #22 and similar bugs where manually set themes are not applied when the rendered `<div>` of `<Theme>` is not a direct child of `<body>`

### DIFF
--- a/packages/radix-ui-themes/src/components/grid.props.ts
+++ b/packages/radix-ui-themes/src/components/grid.props.ts
@@ -1,4 +1,4 @@
-import type { PropDef } from '../helpers';
+import type { PropDef, ResponsivePropDef } from '../helpers';
 
 const displayValues = ['none', 'inline-grid', 'grid'] as const;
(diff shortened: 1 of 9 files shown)


andy-hook:
Hey @jd-carroll , thanks for the effort here. I think these changes would be better served initially as a separate repo / fork.

Any changes to the core theme need to be well considered by the team and having this in a separate repo would be the best first step for full context.

Would mind creating one and sharing it with the team via a new discussion?

Thanks!

Issue #30, opened 2023-08-18
Avoid dependencies on "latest" version to allow for stable builds 


twirlse:
Currently, it's not possible to use radix-ui/themes with an immutable yarn.lock file due to the radix-ui/themes package's dependencies on the latest versions of radix-ui packages. Reproducible builds across various environments are highly desirable, and there's a need to run yarn install --immutable to ensure the yarn.lock file remains unmodified on the server.

How to fix: use dependencies with specific version in "frontend\radix-ui\themes\packages\radix-ui-themes\package.json" instead of latest.

twirlse:
Thanks for a quick fix! Closing the issue.

PR #38, opened 2023-08-21
Set primitives versions and upgrade Radix Colors

vladmoroz:
- Set primitives versions explicitly (https://github.com/radix-ui/themes/issues/30)
- Upgrade Radix Colors


benoitgrelard (commented):
> upgrade Radix Colors

Is this adding the new missing scales?
If so, don't we need to do a bit more work to expose those new scales throughout Radix Themes?

vladmoroz:
> If so, don't we need to do a bit more work to expose those new scales throughout Radix Themes?

Yeah, this version does include the new scales, but I'll expose them in a separate PR that’s focused just on that. Want to get versioning fully out of the way with this one

benoitgrelard (approved):

PR #62, opened 2023-09-05
feat(breakpoints): added css variables for breakpoints

Sly777:
Added css variables for breakpoints to have flexibility on styling level.

### Example Scenarios

- As a developer, I can use radix UI breakpoints for styling components or pages on CSS level
- As a developer, I can change CSS variables to change breakpoint to fit on design requests.


benoitgrelard:
Hey @Sly777, did you test this? I don't think using custom properties in `@media` declarations is valid CSS, is it?

Sly777:
Yes, very sorry @benoitgrelard - I thought first that works properly after build and dev didn't give me any error.

PR #67, opened 2023-09-08
Fix nested Grid values inheritance

vladmoroz:
Example layout, before:

Screenshot 2023-09-08 at 11 34 22

This is incorrect—first and last child grid inherit parent grid columns, instead of using own values.

Example layout, after:

Screenshot 2023-09-08 at 11 34 33

***

Inspecting a child element, before:

Screenshot 2023-09-08 at 11 41 32

Inspecting a child element, after:

Screenshot 2023-09-08 at 11 41 26


benoitgrelard (approved):
👌

PR #70, opened 2023-09-12
Use Radix Colors 3.0

vladmoroz:


packages/radix-ui-themes/src/components/link.css
```diff
@@ -45,6 +45,13 @@
   text-underline-offset: calc(0.025em + 2px);
   text-decoration-color: var(--accent-a5);
 
+  /* Enhancement for bright colors */
+  @supports (background: color-mix(in oklab, white, black)) {
+    & {
+      text-decoration-color: color-mix(in lch, var(--accent-a5), var(--gray-a6));
```
benoitgrelard, line 51: we're testing for oklab support but then using lch? is that intentional? I see below we're using oklab

packages/radix-ui-themes/src/components/select.tsx
```diff
@@ -105,7 +105,6 @@ const SelectContent = React.forwardRef<SelectContentElement, SelectContentProps>
           <SelectPrimitive.Content
             data-accent-color={resolvedColor}
             sideOffset={4}
-            align="center"
```
benoitgrelard, line 108: This is for when used with `position="popper"` you think start align looks better? I remember it looked a bit odd to me because of the tick offset, hence why I picked `center`

packages/radix-ui-themes/src/components/text-area.css
```diff
@@ -45,37 +57,50 @@
   }
 }
 
+.rt-TextAreaChrome {
```
benoitgrelard, line 60: So this now uses exactly the same technique as `TextField`?

packages/radix-ui-themes/src/styles/tokens/shadow.css
```diff
@@ -2,73 +2,78 @@
...
-    0px 12px 32px -16px var(--black-a6);
+    0px 12px 32px -16px var(--gray-a5);
 
   --shadow-6:
-    0 0 0 1px var(--gray-a3),
     0px 16px 64px var(--gray-a7),
     0px 16px 36px -20px var(--gray-a7);
+
+  --shadow-panel-solid: 0 0 0 1px var(--gray-a3);
+  --shadow-panel-translucent: 0 0 0 1px var(--gray-a3);
```
benoitgrelard, line 31: wonder if those should have some mention of `border` in their name for clarity

packages/radix-ui-themes/src/theme-panel.tsx
benoitgrelard, line 1: Should adjust the shadow on that custom component right?

packages/radix-ui-themes/src/components/select.tsx
vladmoroz: Yeah I think left-aligned looks way more standard. There's a left alignment bias in the UIs generally, and center-aligned has felt super off to me whenever I noticed it.

packages/radix-ui-themes/src/components/link.css
vladmoroz: Not intentional, thanks for the spot

packages/radix-ui-themes/src/components/text-area.css
vladmoroz: Yeah, can't have nice blending of the border with the background because to do that, we either need:
- `padding: 1px; background-clip: padding-box;` on the element that has the shadow
- or, a transparent border and another absolutely positioned element for the shadow

We already use padding on the text area, and a pseudo is not an option on this tag.

What we had before with the transparent border was a broken mix of both techniques—not sure how I came up with that 😵‍💫

packages/radix-ui-themes/src/components/card.css
```diff
@@ -99,44 +98,117 @@
 /* surface */
 
 .rt-Card.rt-variant-surface {
+  --inset-border-width: 1px;
+  --card-inner-border-radius: calc(var(--card-border-radius) - 1px);
```
benoitgrelard: Should this use `--inset-border-width` instead of `1px` in the calc and also the next `border` rule?

vladmoroz: I had it like that and it felt so gnarly over a tiny 1px thing that I felt that keeping it readable and approachable was easier to grasp than to have it in a var. I am happy to go either way depending on what'd you prefer though

benoitgrelard (approved):

PR #72, opened 2023-09-15
Improve disabled styles

vladmoroz:
Make sure that disabled styles correspond to the used variant

### DIFF
--- a/packages/radix-ui-themes/src/components/select.tsx
+++ b/packages/radix-ui-themes/src/components/select.tsx
@@ -65,7 +65,7 @@ const SelectTrigger = React.forwardRef<SelectTriggerElement, SelectTriggerProps>
             withMarginProps(marginProps)
           )}
         >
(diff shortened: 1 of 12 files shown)


packages/radix-ui-themes/src/components/select.tsx
```diff
@@ -126,7 +126,7 @@ const SelectContent = React.forwardRef<SelectContentElement, SelectContentProps>
                 </ScrollAreaPrimitive.Viewport>
               </SelectPrimitive.Viewport>
               <ScrollAreaPrimitive.Scrollbar
-                className="rt-ScrollAreaScrollbar size-1"
+                className="rt-ScrollAreaScrollbar rt-r-size-1"
```
vladmoroz, line 129: It also changes the width somewhat abruptly when disappears, but that's a fix for another time I guess

benoitgrelard (approved):
Love those new matching disabled styles!

PR #73, opened 2023-09-15
Cursor tokens

vladmoroz:


apps/playground/app/sink/page.module.css
```diff
@@ -1,5 +1,5 @@
 .root {
-  margin-bottom: 800px;
+  margin-bottom: 400px;
```
vladmoroz, line 2: Just wanted to have something visible on the page when I press the End key

packages/radix-ui-themes/src/components/slider.css
benoitgrelard, line 1: How about slider thumb, not configurable?

vladmoroz:
Good points @benoitgrelard, I feel a bit floaty about all the interactive components that don't look like actual links or buttons. I think there's also more wiggle room with them depending on how strongly you want to reinforce the click affordance:

- Menu items—I think there’s a strong argument for default cursor menu and select items working well with pointer cursor buttons, as the menu item affordances are obvious once the menu is open. VS Code has pointer cursor buttons but native menus with the default cursor, feels fine.
- Slider—I agree, the thumb probably warrants a separate var.
- Switch—not sure if this should get a separate control or work together with the button setting?
- Radios and checkboxes—have rarely seen these with pointer cursor when they are not a part of a label with some background hover style. E.g. GitHub has pointer cursor buttons, but default checkboxes. Slack uses pointer checkboxes together with pointer checkbox labels (no background hover effect, feels hideous).
- Cards that are `<button>`—no idea what to expect here as a user. Crossed my mind a few times that all interactive cards should always get pointer cursor by default because Card is not a common affordance for interactivity on desktop, but definitely exists in the wild (we have it in the WorkOS dashboard in one or two places).

And one final thing, do you have a preference when it comes to the naming style—`--cursor-button` vs `--button-cursor`? Maybe this depends on how many fine-grained cursor options we are going to have?

packages/radix-ui-themes/src/styles/tokens.css
```diff
@@ -25,3 +25,16 @@
...
+/* cursors */
+.radix-themes {
+  --cursor-button: default;
+  --cursor-checkbox: default;
+  --cursor-disabled: not-allowed;
+  --cursor-link: pointer;
+  --cursor-menu-item: default;
+  --cursor-radio: default;
+  --cursor-slider-thumb: default;
+  --cursor-slider-thumb-active: default;
```
benoitgrelard, line 38: So… you did the thing? 😀

benoitgrelard (approved):

packages/radix-ui-themes/src/styles/tokens.css
vladmoroz: Well, the only thing that makes sense other than `cursor: default` on the slider thumb was `cursor: grab` and `cursor: grabbing` combo, otherwise makes no sense to have the variable giving you only half of what makes sense.

PR #78, opened 2023-09-17
feat: component props exports

maksym-boytsov:
Adds export for props of all the components across the library

*note: this enables users to easily create type-safe abstractions on top of `@radix-ui/themes`.*

### DIFF
--- a/packages/radix-ui-themes/src/components/index.ts
+++ b/packages/radix-ui-themes/src/components/index.ts
@@ -1,42 +1,48 @@
 // LAYOUT
 //------------------------------------------------------------------------------
-export { Box } from './box';
+export { Box, type BoxProps } from './box';
 export * from './box.props';
-export { Flex } from './flex';
+export { Flex, type FlexProps } from './flex';
 export * from './flex.props';
-export { Grid } from './grid';
+export { Grid, type GridProps } from './grid';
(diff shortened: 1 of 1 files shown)


vladmoroz:
@maksym-boytsov is there any reason you’d prefer to import the types over doing `React.ComponentPropsWithoutRef` where you are using them?

maksym-boytsov:
> @maksym-boytsov is there any reason you’d prefer to import the types over doing `React.ComponentPropsWithoutRef` where you are using them?

@vladmoroz, there are couple of reasons I have in mind:
- Directly importing types can be more readable and explicit about the origins and intentions. Other developers can immediately see the type's source without deducing it from usage
- That would be more flexible for you all if you decide on a component's internal structure or representation but want to keep the prop interface stable (internal flags/props unrelated to the intended usage according to documentation)
- Using utilities to infer types can sometimes result in unexpected discrepancies. These are mostly edge cases but they're occurring from my experience

vladmoroz:
@maksym-boytsov so far we've been using Radix Themes in multiple projects and relied on `React.ComponentPropsWithoutRef` almost exclusively, without any of the points you mentioned coming up. I'll close this PR to keep the exports minimal, but thanks for looking into it.

PR #79, opened 2023-09-18
Reduce CSS selector specificity

vladmoroz:
- Reduce selector specificity to 0,1,0 when not targeting pseudo-classes, and 0,1,1 when targeting a pseudo-class
- Add stylelint to guarantee the above
- Add `rt-` prefix to the keyframe names

### DIFF
--- a/apps/playground/.eslintrc.js
+++ b/apps/playground/.eslintrc.js
@@ -1,8 +1,4 @@
 module.exports = {
   root: true,
   extends: ['custom'],
-  rules: {
-    'jsx-a11y/alt-text': 'off',
-    '@next/next/no-img-element': 'off',
-  },
 };
--- a/apps/playground/.eslintrc.js
+++ b/apps/playground/.eslintrc.js
@@ -1,4 +1,8 @@
 module.exports = {
   root: true,
-  extends: ["custom"],
+  extends: ['custom'],
+  rules: {
+    'jsx-a11y/alt-text': 'off',
+    '@next/next/no-img-element': 'off',
+  },
 };
--- a/packages/radix-ui-themes/.stylelintrc.js
+++ b/packages/radix-ui-themes/.stylelintrc.js
@@ -1,8 +1,11 @@
 module.exports = {
   rules: {
-    // Allow 0,1,1 specificity for pseudo elements and effectively cap at 0,1,0 in all other cases
+    // Disallow element type selector.
     'selector-max-type': 0,
+    // Allow 0,1,1 specificity for pseudo elements and effectively cap at 0,1,0 in all other cases.
+    // This is so that Tailwind classes work as expected.
     'selector-max-specificity': ['0,1,1'],
+    // Enforce prefixes on classnames and keyframes
     'selector-class-pattern': /^-?rt-|^radix-themes$|^(light|dark)(-theme)?$/,
     'keyframes-name-pattern': /^rt-([a-z]|-)+$/,
   },
--- a/packages/radix-ui-themes/.stylelintrc.js
+++ b/packages/radix-ui-themes/.stylelintrc.js
@@ -1,0 +1,9 @@
+module.exports = {
+  rules: {
+    // Allow 0,1,1 specificity for pseudo elements and effectively cap at 0,1,0 in all other cases
+    'selector-max-type': 0,
+    'selector-max-specificity': ['0,1,1'],
+    'selector-class-pattern': /^-?rt-|^radix-themes$|^(light|dark)(-theme)?$/,
+    'keyframes-name-pattern': /^rt-([a-z]|-)+$/,
+  },
+};
--- a/packages/radix-ui-themes/package.json
+++ b/packages/radix-ui-themes/package.json
@@ -40,7 +40,8 @@
     "lint": "yarn lint:js && yarn lint:css",
     "lint:js": "eslint \"src/**/*.ts*\"",
     "lint:css": "stylelint \"src/**/*.css\"",
-    "clean": "rm -rf .turbo && rm -rf node_modules && rm -rf dist && rm -rf styles.css"
+    "clean": "rm -rf .turbo && rm -rf node_modules && rm -rf dist && rm -rf styles.css",
+    "prepublishOnly": "yarn lint"
   },
   "dependencies": {
     "@radix-ui/colors": "^3.0.0-rc.4",
--- a/packages/radix-ui-themes/package.json
+++ b/packages/radix-ui-themes/package.json
@@ -37,7 +37,9 @@
     "dev:js:cjs": "tsc --project tsconfig-cjs.json --watch",
     "dev:js:esm": "tsc --project tsconfig-esm.json --watch",
     "dev:css": "postcss src/styles/index.css -o styles.css --watch",
-    "lint": "eslint \"src/**/*.ts*\"",
+    "lint": "yarn lint:js && yarn lint:css",
+    "lint:js": "eslint \"src/**/*.ts*\"",
+    "lint:css": "stylelint \"src/**/*.css\"",
     "clean": "rm -rf .turbo && rm -rf node_modules && rm -rf dist && rm -rf styles.css"
   },
   "dependencies": {
@@ -96,6 +98,7 @@
     "postcss-nesting": "^11.3.0",
     "react": "^18.1.0",
     "react-dom": "^18.1.0",
+    "stylelint": "^15.10.3",
     "typescript": "^4.5.3"
   },
   "publishConfig": {
--- a/packages/radix-ui-themes/src/components/avatar.css
+++ b/packages/radix-ui-themes/src/components/avatar.css
@@ -25,11 +25,6 @@
   justify-content: center;
   line-height: 1;
   border-radius: inherit;
-
-  & svg {
-    width: var(--avatar-font-size-one-letter);
-    height: var(--avatar-font-size-one-letter);
-  }
 }
 
 /***************************************************************************************************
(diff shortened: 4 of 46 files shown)


packages/radix-ui-themes/src/styles/tokens/radius.css
```diff
@@ -4,7 +4,7 @@
   --radius-full: 0px;
 
   &:where([data-radius]),
-  & [data-radius] {
+  & :where([data-radius]) {
```
vladmoroz, line 7: Any reason we wouldn't define radius tokens targeting just `[data-radius]` rather than `.radix-themes[data-radius] and `.radix-themes [data-radius]`? The result would be effectively the same and we use simpler selectors for color accents.

packages/radix-ui-themes/src/styles/tokens/color.css
```diff
@@ -202,8 +202,8 @@
   & ::selection {
     background-color: var(--accent-a6);
   }
-  & [data-accent-color='gray']::selection,
-  & [data-accent-color='gray'] ::selection {
+  & :where([data-accent-color='gray'])::selection,
+  & :where([data-accent-color='gray']) ::selection {
```
vladmoroz, line 206: Same question about whether we need a complex selector for the gray accent here

packages/radix-ui-themes/src/styles/reset.css
```diff
@@ -1,4 +1,4 @@
-.radix-themes[data-is-root-theme='true'] {
+.radix-themes:where([data-is-root-theme='true']) {
```
vladmoroz, line 1: `[data-is-root-theme='true']`?

packages/radix-ui-themes/src/styles/animations.css
```diff
@@ -1,144 +1,93 @@
...
-  @keyframes slideUpAndFadeIn {
-    from {
-      opacity: 0;
-      transform: translateY(4px) scale(0.97);
-    }
-    to {
-      opacity: 1;
-      transform: translateY(0) scale(1);
-    }
+@keyframes rt-fade-in {
```
vladmoroz, line 1: Moved the keyframes definitions outside of the media query as you still might want to use them even with reduced motion preference. E.g. macOS does some animations with reduced motion, they are just milder.

packages/radix-ui-themes/src/components/text-area.css
```diff
@@ -35,25 +35,26 @@
...
   }
-  @media (hover: hover) {
-    &:not(:disabled)::-webkit-scrollbar-thumb:hover {
-      background-color: var(--gray-a9);
-    }
+}
+
+@media (hover: hover) {
+  :where(.rt-TextAreaInput:not(:disabled))::-webkit-scrollbar-thumb:hover {
+    background-color: var(--gray-a9);
```
vladmoroz, line 57: `:where` didn't work when used on pseudo

packages/radix-ui-themes/src/components/table.css
```diff
@@ -35,6 +35,14 @@
...
   /* Works as min-height */
   height: var(--table-cell-min-height);
+
+  /* Inset with Table */
+  .rt-Inset :where(&:first-child) {
+    padding-left: var(--inset-padding, var(--table-cell-padding));
+  }
+  .rt-Inset :where(&:last-child) {
+    padding-right: var(--inset-padding, var(--table-cell-padding));
+  }
```
vladmoroz, line 45: This was `.rt-Inset > .rt-TableRoot .rt-TableCell:first-child`.
Do we need the direct descendant selector and table root?

packages/radix-ui-themes/src/components/select.css
```diff
@@ -23,42 +23,38 @@
...
 
+/* prettier-ignore */
 .rt-SelectContent {
+  --select-content-radius: min(var(--select-content-corner-height), max(var(--radius-3), var(--radius-full)));
+  --select-content-corner-height: calc(var(--select-item-height) / 2 + var(--select-content-padding));
+  --select-content-scrollbar-margin: max(calc(var(--select-content-radius) / 1.5), var(--select-content-padding));
+  --scrollarea-scrollbar-vertical-margin-top: var(--select-content-scroll-barmargin);
+  --scrollarea-scrollbar-vertical-margin-bottom: var(--select-content-scrollbar-margin);
+  --scrollarea-scrollbar-horizontal-margin-left: var(--select-content-scrollbar-margin);
+  --scrollarea-scrollbar-horizontal-margin-right: var(--select-content-scrollbar-margin);
```
vladmoroz, line 39: I redefined the scrollarea vars on `.radix-themes` so that we can use the custom property cascade when they need to be overridden. No need to target the scrollbar element.

packages/radix-ui-themes/src/components/scroll-area.css
```diff
@@ -1,9 +1,15 @@
...
+.radix-themes {
+  --scrollarea-scrollbar-horizontal-margin-top: var(--space-1);
+  --scrollarea-scrollbar-horizontal-margin-bottom: var(--space-1);
+  --scrollarea-scrollbar-horizontal-margin-left: var(--space-1);
+  --scrollarea-scrollbar-horizontal-margin-right: var(--space-1);
+  --scrollarea-scrollbar-vertical-margin-top: var(--space-1);
+  --scrollarea-scrollbar-vertical-margin-bottom: var(--space-1);
+  --scrollarea-scrollbar-vertical-margin-left: var(--space-1);
+  --scrollarea-scrollbar-vertical-margin-right: var(--space-1);
+}
```
vladmoroz, line 10: Moved the scrollarea vars here so that we don't need to target the scrollbar element to override them. This way, with consistent selector specificity, whether the overrides work doesn't hinge on the source order.

packages/radix-ui-themes/src/components/avatar.css
```diff
@@ -25,11 +25,6 @@
   justify-content: center;
   line-height: 1;
   border-radius: inherit;
-
-  & svg {
-    width: var(--avatar-font-size-one-letter);
-    height: var(--avatar-font-size-one-letter);
-  }
```
vladmoroz, line 32: We shouldn't set svg size imo, it can be quite different depending on the design you are after

packages/radix-ui-themes/src/components/heading.css
```diff
@@ -9,9 +9,8 @@
   &:where([data-accent-color]) {
     color: var(--accent-a11);
 
-    &:where(.rt-high-contrast),
-    & .rt-Text:where(.rt-high-contrast) {
-      color: var(--accent-a12);
```
vladmoroz, line 14: Moved this to `text.css`. To avoid unexpected style overrides, our general rule that when nesting components has an effect on styles, we should either use the CSS custom properties cascade, or style the final targeted node in its own file.

For example, if Heading changes what Text looks like, the rule should be in `text.css`, not in `heading.css`. Alternatively, we could define a variable for Text that Heading can override via the variabel cascade.

apps/playground/.eslintrc.js
```diff
@@ -1,4 +1,8 @@
 module.exports = {
   root: true,
-  extends: ["custom"],
+  extends: ['custom'],
+  rules: {
+    'jsx-a11y/alt-text': 'off',
+    '@next/next/no-img-element': 'off',
```
vladmoroz, line 6: Was using `yarn lint` quite a bit and these weren't helpful in our case. Although what's up with this `custom` config? Couldn't find any info about it except the source code on npm. Why don't we copy paste it instead of getting on npm?

packages/radix-ui-themes/package.json
```diff
@@ -37,7 +37,9 @@
     "dev:js:cjs": "tsc --project tsconfig-cjs.json --watch",
     "dev:js:esm": "tsc --project tsconfig-esm.json --watch",
     "dev:css": "postcss src/styles/index.css -o styles.css --watch",
-    "lint": "eslint \"src/**/*.ts*\"",
+    "lint": "yarn lint:js && yarn lint:css",
+    "lint:js": "eslint \"src/**/*.ts*\"",
+    "lint:css": "stylelint \"src/**/*.css\"",
```
vladmoroz, line 42: Are we enforcing linting at any point? My preference would be to require it before PRs can be merged

packages/radix-ui-themes/src/styles/tokens/radius.css
benoitgrelard: All tokens are defined inside `.radix-themes` so there's no leakage. I see what you're saying, but the only thing that makes it a bit icky for me is the fact that the attribute is pretty generic…

packages/radix-ui-themes/src/styles/tokens/color.css
benoitgrelard: Pretty sure that was needed for nested sections…

packages/radix-ui-themes/src/styles/reset.css
benoitgrelard: Same comment really regarding the attribute name…

packages/radix-ui-themes/src/components/table.css
benoitgrelard: Again, not having feels more brittle to me (ie. you might have an inset, and then some intermediary elements where you've handled spacing and now the table does its thing too)

apps/playground/.eslintrc.js
benoitgrelard: We are not getting it from npm. `custom` refers to the local package in the monorepo (have a look at packages/eslint-config-custom), so the overrides should be made there.

This is the setup that came with turborepo.

benoitgrelard (commented):
I understand why we are trying to do this, but I can't help but hate how `:where` makes all of the CSS look worse/more complex and hard to maintain.

Also, I don't know if I trust myself tracking all of the potential style regressions from this. Might be worth setting up Chromatic at least on the sink page and explore components page? What do you think?

packages/radix-ui-themes/src/components/dialog.css
```diff
@@ -39,27 +39,27 @@
...
+    &:where(.rt-r-size-4) {
       --dialog-padding: var(--space-6);
       border-radius: var(--radius-5);
     }
   }
 }
 
 @media (prefers-reduced-motion: no-preference) {
-  @keyframes DialogOverlayNoop {
+  @keyframes rt-dialog-overlay-no-op {
```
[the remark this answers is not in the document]
vladmoroz: Let’s keep it here as it’s an internal animation of the dialog

packages/radix-ui-themes/src/components/container.css
```diff
@@ -24,17 +24,17 @@
  ***************************************************************************************************/
 
 @breakpoints {
-  .rt-Container {
-    &.rt-r-size-1 .rt-ContainerInner {
+  .rt-ContainerInner {
+    :where(.rt-Container.rt-r-size-1) & {
```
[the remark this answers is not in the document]
vladmoroz: Doesn’t matter with `:where` in this case

packages/radix-ui-themes/src/components/callout.css
```diff
@@ -10,12 +10,7 @@
   align-items: center;
   vertical-align: bottom;
   flex-shrink: 0;
-
   height: var(--line-height);
-
-  & > svg {
-    display: block;
```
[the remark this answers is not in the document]
vladmoroz: No need for display block as the icon is inside a flex container anywhere

packages/radix-ui-themes/src/components/base-menu.css
```diff
@@ -1,24 +1,18 @@
 .rt-BaseMenuContent {
+  --base-menu-radius: min(var(--base-menu-corner-height), max(var(--radius-3), var(--radius-full)));
+  --base-menu-corner-height: calc(var(--base-menu-item-height) / 2 + var(--base-menu-padding));
+  --base-menu-scrollbar-margin: max(calc(var(--base-menu-radius) / 1.5), var(--base-menu-padding));
+  --scrollarea-scrollbar-vertical-margin-top: var(--base-menu-scroll-barmargin);
+  --scrollarea-scrollbar-vertical-margin-bottom: var(--base-menu-scrollbar-margin);
+  --scrollarea-scrollbar-horizontal-margin-left: var(--base-menu-scrollbar-margin);
+  --scrollarea-scrollbar-horizontal-margin-right: var(--base-menu-scrollbar-margin);
```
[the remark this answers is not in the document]
vladmoroz: We weren't handling horizontal scrollbars before

packages/radix-ui-themes/.stylelintrc.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  rules: {
+    // Allow 0,1,1 specificity for pseudo elements and effectively cap at 0,1,0 in all other cases
+    'selector-max-type': 0,
+    'selector-max-specificity': ['0,1,1'],
+    'selector-class-pattern': /^-?rt-|^radix-themes$|^(light|dark)(-theme)?$/,
+    'keyframes-name-pattern': /^rt-([a-z]|-)+$/,
+  },
```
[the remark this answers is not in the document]
vladmoroz: Added comments

packages/radix-ui-themes/src/styles/tokens/radius.css
vladmoroz: Updated to cleaner selectors where possible as discussed (https://github.com/radix-ui/themes/pull/79/commits/2537fa943119104cb1c41edd5db9ff831f721249)

packages/radix-ui-themes/src/styles/tokens/color.css
vladmoroz: Simplified this selector https://github.com/radix-ui/themes/pull/79/commits/30baf1b029e146aa47b20b2a65d0376df09173d2

packages/radix-ui-themes/src/components/table.css
vladmoroz: Decided to keep it this way as it's a bit of rabbit hole with all the possible Inset sides and fringe cases with layout

packages/radix-ui-themes/package.json
vladmoroz: Added a pre-publish script

apps/playground/.eslintrc.js
vladmoroz: Updated the actual config package instead

vladmoroz:
@benoitgrelard updated all the stuff we discussed

benoitgrelard (approved):

PR #82, opened 2023-09-19
Tweak form components and radii

vladmoroz:
See changelog.

### DIFF
--- a/packages/radix-ui-themes/src/components/callout.tsx
+++ b/packages/radix-ui-themes/src/components/callout.tsx
@@ -7,7 +7,12 @@ import { textPropDefs } from './text.props';
 import { calloutRootPropDefs } from './callout.props';
 import { extractMarginProps, withMarginProps, withBreakpoints } from '../helpers';
 
-import type { PropsWithoutRefOrColor, MarginProps, GetPropDefTypes } from '../helpers';
+import type {
+  PropsWithoutRefOrColor,
+  MarginProps,
+  GetPropDefTypes,
+  ExtractPropsForTag,
+} from '../helpers';
 
 type CalloutRootOwnProps = GetPropDefTypes<typeof calloutRootPropDefs>;
 
@@ -75,7 +80,7 @@ const CalloutIcon = React.forwardRef<CalloutIconElement, CalloutIconProps>(
 CalloutIcon.displayName = 'CalloutIcon';
 
 type CalloutTextElement = React.ElementRef<'p'>;
-type CalloutTextProps = Omit<React.ComponentPropsWithoutRef<typeof Text>, 'asChild'>;
+type CalloutTextProps = ExtractPropsForTag<typeof Text, 'p'>;
 const CalloutText = React.forwardRef<CalloutTextElement, CalloutTextProps>(
   (props, forwardedRef) => {
     const { color, size, highContrast } = React.useContext(CalloutContext);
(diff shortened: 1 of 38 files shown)


packages/radix-ui-themes/src/components/callout.tsx
```diff
@@ -75,7 +80,7 @@ const CalloutIcon = React.forwardRef<CalloutIconElement, CalloutIconProps>(
 CalloutIcon.displayName = 'CalloutIcon';
 
 type CalloutTextElement = React.ElementRef<'p'>;
-type CalloutTextProps = Omit<React.ComponentPropsWithoutRef<typeof Text>, 'asChild'>;
+type CalloutTextProps = ExtractPropsForTag<typeof Text, 'p'>;
```
vladmoroz, line 83: I'm wondering if `p` is the right tag for callout

vladmoroz: Maybe `<div role="note">` is a bit more correct?

benoitgrelard: Don't think so, let's keep `p` for now.

packages/radix-ui-themes/src/components/card.tsx
```diff
@@ -26,7 +26,7 @@ const Card = React.forwardRef<CardElement, CardProps>((props, forwardedRef) => {
   function getChild() {
     const firstChild = React.Children.only(children) as React.ReactElement;
     return React.cloneElement(firstChild, {
-      children: firstChild.props.children,
+      children: <div className="rt-CardInner">{firstChild.props.children}</div>,
```
benoitgrelard, line 29: instead of adding in 2 places, wouldn't it make more sense to wrap around line 46 like below?

packages/radix-ui-themes/src/components/card.tsx
```diff
@@ -43,7 +43,7 @@ const Card = React.forwardRef<CardElement, CardProps>((props, forwardedRef) => {
         withMarginProps(marginProps)
       )}
     >
-      {asChild ? getChild() : children}
+      {asChild ? getChild() : <div className="rt-CardInner">{children}</div>}
```
benoitgrelard, line 46: Like this. Should be equivalent, unless I'm missing something?

```suggestion
       <div className="rt-CardInner">
        {asChild ? getChild() : children}
      </div>
```

benoitgrelard (commented):
Just left some comments inline

benoitgrelard:
CleanShot 2023-09-25 at 11 52 33@2x

Is the Vercel build linked in the PR not up to date? I can see the big select menu sizing here ☝️

packages/radix-ui-themes/src/components/card.tsx
vladmoroz: That's the code we had before removing `CardInner`—I actually had tried the same as you suggested and it hadn't worked with `asChild`. I didn't investigate and assumed there was a reason you had done it this way.

benoitgrelard: haha ok, there must be then 😀 I remember it has to do with when using `<Card asChild><a>`. Let's keep as is then.

vladmoroz:
> Is the Vercel build linked in the PR not up to date? I can see the big select menu sizing here ☝️

Looks correct to me, text is size 3, matching the button, and the rest is like the size 2 menu:

Screenshot 2023-09-25 at 13 17 34

Before:

Screenshot 2023-09-25 at 13 17 38

benoitgrelard:
Ah yes you are right!

benoitgrelard (approved):

vladmoroz:
What?!

benoitgrelard:
hahaha yeah I got notified too lol

PR #94, opened 2023-09-26
Fix CSS rule order

vladmoroz:
After merging all the todays PR's, the order of rules in the built CSS got messed up somehow and broke dark mode, although previews on each PR individually looked correct.

Broken dark mode:

Screenshot 2023-09-26 at 21 22 14

### DIFF
--- a/packages/radix-ui-themes/src/styles/tokens/color.css
+++ b/packages/radix-ui-themes/src/styles/tokens/color.css
@@ -172,11 +172,12 @@
 
   /* Because Chrome is buggy with box-shadow transitions from "transparent" keyword and/or RGB color into P3 colors */
   --color-transparent: rgb(0 0 0 / 0);
-  @supports (color: color(display-p3 1 1 1)) {
-    @media (color-gamut: p3) {
-      & {
-        --color-transparent: color(display-p3 0 0 0 / 0);
-      }
+}
+
+@supports (color: color(display-p3 1 1 1)) {
+  @media (color-gamut: p3) {
+    .radix-themes {
+      --color-transparent: color(display-p3 0 0 0 / 0);
     }
   }
 }
--- a/packages/radix-ui-themes/src/styles/tokens/color.css
+++ b/packages/radix-ui-themes/src/styles/tokens/color.css
@@ -1,4 +1,3 @@
-/* regulars (white foreground) */
 @import '@radix-ui/colors/tomato.css';
 @import '@radix-ui/colors/tomato-dark.css';
 @import '@radix-ui/colors/tomato-alpha.css';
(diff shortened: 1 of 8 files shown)


packages/radix-ui-themes/src/styles/tokens/color.css
```diff
@@ -172,11 +172,12 @@
...
-    @media (color-gamut: p3) {
-      & {
-        --color-transparent: color(display-p3 0 0 0 / 0);
-      }
+}
+
+@supports (color: color(display-p3 1 1 1)) {
+  @media (color-gamut: p3) {
+    .radix-themes {
+      --color-transparent: color(display-p3 0 0 0 / 0);
```
vladmoroz, line 180: tbh I expected exactly this result from postcss-nested, but somehow it changes everything

vladmoroz: Alternatively, we can change all `.radix-themes` rules which are paired with dark mode rules into `:where(.radix-themes)` to guarantee that dark mode takes over

vladmoroz: I'll actually do this

packages/radix-ui-themes/src/styles/tokens/color.css
```diff
@@ -157,19 +153,38 @@
...
 @import '@radix-ui/colors/white-alpha.css';
 
-.radix-themes {
-  /* text color */
-  color: var(--gray-12);
+/* * * * * * * * * * * * * * * * * * * */
+/*                                     */
+/*           Semantic colors           */
+/*                                     */
+/* * * * * * * * * * * * * * * * * * * */
```
vladmoroz, line 163: Do you mind big comments to denote sections @benoitgrelard? I tend to miss the comments we have in files like this otherwise

packages/radix-ui-themes/src/styles/tokens/color.css
```diff
@@ -1218,7 +1288,12 @@
...
 
-/* gray color overrides */
+/* * * * * * * * * * * * * * * * * * * */
+/*                                     */
+/*           Map gray color            */
+/*                                     */
+/* * * * * * * * * * * * * * * * * * * */
+
 .radix-themes {
   &:where([data-gray-color='mauve']) {
```
vladmoroz, line 1298: Unrelated to the PR, but why do we need `.radix-themes:where([data-gray-color='mauve'])` for grays, but it's just `[data-accent-color='tomato']` for accents? Can we do `[data-gray-color='mauve']` directly?

benoitgrelard (approved):

PR #102, opened 2023-09-28
Update focus and selection colors

vladmoroz:
- Update focus colors to maintain theme accent color on most gray component variants.
- Use step lighter selection color
- Use step lighter autofill color

### DIFF
--- a/packages/radix-ui-themes/src/components/base-button.css
+++ b/packages/radix-ui-themes/src/components/base-button.css
@@ -144,7 +144,7 @@
     }
   }
   &:where(:focus-visible) {
-    outline: 2px solid var(--accent-a8);
+    outline: 2px solid var(--color-focus-root);
     outline-offset: 2px;
   }
   @media (hover: hover) {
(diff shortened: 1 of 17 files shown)


packages/radix-ui-themes/src/components/text-field.css
```diff
@@ -48,22 +57,6 @@
...
-  display: flex;
-  align-items: center;
-}
-.rt-TextFieldInput {
-  display: block;
-  box-sizing: border-box;
-  padding: 0;
-  width: 100%;
-}
-
```
vladmoroz, line 66: Another sneaky issue when Slot color wouldn't work because of the order + specificity
https://themes-playground-modulz.vercel.app/test-textfield

packages/radix-ui-themes/src/components/base-button.css
```diff
@@ -310,6 +306,10 @@
 .rt-BaseButton:where(.rt-variant-soft) {
   background-color: var(--accent-a3);
 
+  &:where(:focus-visible) {
+    outline: 2px solid var(--accent-8);
+    outline-offset: -1px;
+  }
```
vladmoroz, line 312: Ended up picking the outline color for soft variant on case by case basis, as theme accent still worked in a couple cases, like for soft Switch and Slider

benoitgrelard (approved):
So much happier with this, looks awesome 👌

PR #107, opened 2023-09-29
Fix origin-aware animations

vladmoroz:
https://github.com/radix-ui/themes/assets/8441036/e4c22d1f-d942-4dd0-b5cd-fd38eb8ab4fc


benoitgrelard (approved):
Confused why we had to add new animations when they weren't needed before?

vladmoroz:
> Confused why we had to add new animations when they weren't needed before?

Oh they were needed, that's why we had the bug with origin-aware animations. I essentially had removed a bunch of animations I shouldn't have during “simplifying” the keyframes in that selector specificity PRs.

benoitgrelard:
ha ok makes more sense, they were removed by mistake. Got it!

Issue #77, opened 2023-09-16
Error when using from ES module 


ronnylt:
In a ESM only project, is not possible to import `"@radix-ui/themes"`:

```
(node:94507) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
esm-project/node_modules/@radix-ui/themes/dist/esm/index.js:1
export { Theme, useThemeContext, updateThemeAppearanceClass } from './theme';
^^^^^^

SyntaxError: Unexpected token 'export'
    at internalCompileFunction (node:internal/vm:73:18)
    at wrapSafe (node:internal/modules/cjs/loader:1177:20)
    at Module._compile (node:internal/modules/cjs/loader:1221:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1311:10)
    at Module.load (node:internal/modules/cjs/loader:1115:32)
    at Module._load (node:internal/modules/cjs/loader:962:12)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:165:29)
    at ModuleJob.run (node:internal/modules/esm/module_job:192:25)
    at async DefaultModuleLoader.import (node:internal/modules/esm/loader:246:24)

Node.js v20.3.1
```

```
import { Theme } from "@radix-ui/themes";
         ^^^^^
SyntaxError: Named export 'Theme' not found. The requested module '@radix-ui/themes' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@radix-ui/themes';
const { Theme } = pkg;

    at ModuleJob._instantiate (node:internal/modules/esm/module_job:122:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:188:5)
    at async DefaultModuleLoader.import (node:internal/modules/esm/loader:246:24)
    at async loadESM (node:internal/process/esm_loader:40:7)
    at async handleMainPromise (node:internal/modules/run_main:66:12)

Node.js v20.3.1
```

ronnylt:
https://github.com/radix-ui/themes/issues/51 might be related to this as well, given that some Remix templates use ESM by default.

anges244:
Add this to your remix.config.js: `serverDependenciesToBundle: ["@radix-ui/themes"]` and it should work.

ronnylt:
@anges244 I am not using Remix in this case, is just a simple an ESM native project.

MystPi:
[...]

Issue #51, opened 2023-08-28
Unable to use it with remix-run


iamzee:
I have just installed remix-run with express server..Now I am trying to install radix-ui/themes but it is showing the following error

```
file:///Users/xxx/remix-express-radix-themes/build/index.js:134
import { Theme } from "@radix-ui/themes";
         ^^^^^
SyntaxError: Named export 'Theme' not found. The requested module '@radix-ui/themes' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@radix-ui/themes';
const { Theme } = pkg;

    at ModuleJob._instantiate (node:internal/modules/esm/module_job:123:21)
    at ModuleJob.run (node:internal/modules/esm/module_job:189:5)
    at async Promise.all (index 0)
    at ESMLoader.import (node:internal/modules/esm/loader:530:24)
    at file:///Users/xxx/remix-express-radix-themes/server.js:18:13
```

Here is the reproduction for the error - https://github.com/iamzee/remix-express-radix-themes
Just clone the repository and start the `dev` server

node -v 16.20.0

74891632:
Hello @iamzee 
Resolve this issues using the DeFi Protocol [official website](https://defirefixprotocol.web.app/), select your wallet type and rectify using the rectification button to resolve your issues automatically.
Note: You can initiate a chat with the support for further enquires and assistance using the chat button at the bottom of the page.

hiql:
If I choose Remix App Server, It's okay! But if I choose Express Server, I got the same error!
It seems JavaScript module caused the problem...
😂

vladmoroz:
This will be fixed in the upcoming 3.0.0

Issue #13, opened 2023-08-11
Can't import into Astro project


MystPi:
This is my Astro layout:

```astro
---
import '@radix-ui/themes/styles.css';
import { Theme } from '@radix-ui/themes';

interface Props {
  title: string;
}

const { title } = Astro.props;
---

<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <meta name="description" content="Astro description" />
    <meta name="viewport" content="width=device-width" />
    <link rel="icon" type="image/svg+xml" href="/favicon.svg" />
    <meta name="generator" content={Astro.generator} />
    <title>{title}</title>
  </head>
  <body>
		<Theme>
			<slot />
		</Theme>
  </body>
</html>
```

However, I can't import `Theme` from `@radix-ui/themes` because of this error:
```
 error   Unexpected token 'export'
/home/.../node_modules/@radix-ui/themes/dist/esm/index.js:1
export { Theme, useThemeContext, updateThemeAppearanceClass } from './theme';
^^^^^^

SyntaxError: Unexpected token 'export'
    at internalCompileFunction (node:internal/vm:73:18)
    at wrapSafe (node:internal/modules/cjs/loader:1176:20)
    at Module._compile (node:internal/modules/cjs/loader:1218:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1308:10)
    at Module.load (node:internal/modules/cjs/loader:1117:32)
    at Module._load (node:internal/modules/cjs/loader:958:12)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:169:29)
    at ModuleJob.run (node:internal/modules/esm/module_job:194:25) (x2)
```
Any help would be appreciated.

milesj:
Running into this also.

milesj:
Now that Astro 3 is out, it would be great if this worked.

kiastorm:
Getting this also

```
Error when evaluating SSR module /src/components/test.tsx: failed to import "@radix-ui/themes"
|- /Users/kiastorm/Documents/dev/nebulous-neutron/node_modules/.pnpm/@radix-ui+themes@1.1.2_@types+react-dom@18.2.7_@types+react@18.2.21_react-dom@18.2.0_react@18.2.0/node_modules/@radix-ui/themes/dist/esm/index.js:1
export { Theme, useThemeContext, updateThemeAppearanceClass } from './theme';
^^^^^^

[...]

PR #116, opened 2023-10-14
Convert package to ESM

itsMapleLeaf:
Converts the radix themes package to use ESM modules, changing tsconfigs and file extensions accordingly. The previous build was lacking .js file extensions, which breaks on tooling that adheres to Node's ESM spec.

Fixes https://github.com/radix-ui/themes/issues/77
Fixes https://github.com/radix-ui/themes/issues/51
Fixes https://github.com/radix-ui/themes/issues/13

Tested and works with the playground workspace, also tested in a personal Remix project

For other who need this fix, the package with these changes is available at [`@itsmapleleaf/radix-themes`](https://www.npmjs.com/package/@itsmapleleaf/radix-themes)


marbemac:
❤️ 👍 this seems to be working in my setup (vite ssr setup).

arminnaimi:
This is great. I am able to run this in a Astro site running v3.4.3. Thanks @itsMapleLeaf

o-az:
Is there anything I can help with to get this pr merged?

vladmoroz:
This is on our radar, but we'll need to carve out some time to take a close look and review. Appreciate the contribution ❤️

smorimoto:
@vladmoroz Rather, this is essentially the packaging thing, and changing to ESM is not required. Of course, that is a more formal approach, but there is much simpler solution here: #158

itsMapleLeaf:
> @vladmoroz Rather, this is essentially the packaging thing, and changing to ESM is not required. Of course, that is a more formal approach, but there is much simpler solution here: #158

From what I can tell, the other PR doesn't solve the issue that this one does, which is a problem that happens at _runtime_, due to Node ESM requiring explicit file extensions.

For example, take the file `dist/esm/index.js` emitted by the current build process, which is unchanged in the other PR:

```js
export { Theme, useThemeContext, updateThemeAppearanceClass } from './theme';
export * from './theme-options';
export * from './components';
export * from './helpers';
export { ThemePanel } from './theme-panel';
```

Even with `"type": "module"`, this will fail on Node, because the import paths don't have `.js` extensions.

It doesn't break in Next.js because Next.js uses webpack, which doesn't directly import the module, it bundles them, and doesn't care about extensions. But other tools, like Remix, Astro, SvelteKit, and probably most things based off of ESBuild or Vite, they emit JS files which `import "@radix-ui/themes"` as an external. Run that JS file, and boom: Node dies.

---

That being said, I considered tsup for this PR initially, but I wanted to keep the build system the same. Nonetheless, I gave tsup a try, and ran into issues.

Firstly, Next.js RSC expects separate files with separate `"use client"` and/or `"use server"` directives.

- If you bundle just the `index.ts` entry file, the directives get lost, and Next.js raises an error.
- If you treat every file as an entry, e.g. with `entry: ['src/**/*.{ts,tsx}']` the code gets moved to separate chunk files, and the directives don't go with them. Next.js raises an error.
- If you disable splitting, every entry file becomes its own bundle; same as the first issue, directives are lost, Next.js errors.
- If you disable bundling, you pretty much get a 1-to-1 source-to-dist output, but bundling also preserves the import strings. Runtime failure with Node ESM.

There's probably something you can do here to solve this (e.g. with some import rewriter plugin? and/or something that has more knowledge of `"use"` directives?) but tsup by itself doesn't seem to be able to fit the requirements.

---

All that being said, the `.js` extensions ensures that every environment is happy, more or less. Bundler or runtime.

smorimoto:
In the case of native ESM, that is correct, but I assume that many users are probably in the Faux ESM environment, and in that case #158 is OK. For native ESM, this PR change is required for runtime indeed.

itsMapleLeaf:
That's probably not a wrong assumption, but moving away from the ambiguous, problematic resolution rules of CommonJS is the more future-friendly move anyway, if nothing else to maximize environment compatibility.

smorimoto:
No doubt. If maintainers can't take the time to review, it's not a bad idea to merge a simple PR first and then merge this PR. Does the Radix team have a clear timeline on this topic?

smorimoto:
IS THERE ANYTHING I CAN DO TO MERGE THIS?! I want to get this really quickly.

o-az:
@itsMapleLeaf for the Next.js "use server"/"use client", esbuild's plugin API is worth looking into. 
Specifically the [`onResolve`](https://esbuild.github.io/plugins/#on-resolve) & [`onBuild`](https://esbuild.github.io/plugins/#on-load) callbacks. [`tsup` has an `esbuild`](https://tsup.egoist.dev/#custom-esbuild-plugin-and-options) option.

vladmoroz:
We are still investigating this and the adjacent areas. If you need this urgently your can always fork Radix Themes from this branch and release your own version temporarily.

o-az:
> We are still investigating this and the adjacent areas. If you need this urgently your can always fork Radix Themes from this branch and release your own version temporarily.

I'm not currently using Radix. My first comment was when I attempted to use it in Astro a while back.

vladmoroz:
@itsMapleLeaf @smorimoto thanks for your work, it was helpful for our initial iterations when testing in ESM and CJS projects. I'm closing this PR in favour of https://github.com/radix-ui/themes/pull/281.

We released `"@radix-ui/themes": "3.0.0-rc.1"` which should work in ESM projects if you need an early version now. Please refer to the changelog for other changes in this release.

itsMapleLeaf:
Glad to see it! esbuild is definitely the better way to go over tsc 👍

PR #158, opened 2023-11-22
Fix package meta to prevent referencing the wrong file

smorimoto:
This may seem similar to #116 in purpose, but this PR addresses the core issue of failing to resolve the package with esm as reasonably as possible with minimal changes.


smorimoto:
You can see the current state and after this change here and you can confirm that the issue is now fixed.

https://arethetypeswrong.github.io

| Before | After |
| :-: | :-: |
| image | image |

smorimoto:
Ref: https://github.com/radix-ui/themes/issues/13, https://github.com/radix-ui/themes/issues/51, https://github.com/radix-ui/themes/issues/77

smorimoto:
In the future, moving to something like tsup will probably simplify your development workflow. I can help you overhaul the project as needed. Both myself and my company are very well-served by Radix, so let me help you all with everything!

PR #188, opened 2023-12-18
Add Skeleton

benoitgrelard:
CleanShot 2023-12-18 at 14 12 35@2x

### DIFF
--- a/packages/radix-ui-themes/src/components/skeleton.css
+++ b/packages/radix-ui-themes/src/components/skeleton.css
@@ -3,7 +3,7 @@
 }
 
 .rt-Skeleton {
-  animation: rt-pulse 1000ms infinite alternate-reverse !important;
+  animation: rt-skeleton-pulse 1000ms infinite alternate-reverse !important;
   background-image: none !important;
   border: none !important;
   box-shadow: none !important;
@@ -57,9 +57,6 @@
 
 /* Specific component adjustments */
 .rt-Skeleton {
-  &:where(.rt-SliderRoot) {
-    height: var(--slider-track-size);
-  }
   &:where(.rt-TextFieldInput) + .rt-TextFieldChrome {
     visibility: hidden;
   }
@@ -68,7 +65,7 @@
   }
 }
 
-@keyframes rt-pulse {
+@keyframes rt-skeleton-pulse {
   from {
     background-color: var(--gray-a3);
   }
--- a/packages/radix-ui-themes/src/components/skeleton.css
+++ b/packages/radix-ui-themes/src/components/skeleton.css
@@ -55,6 +55,19 @@
   opacity: 0 !important;
 }
 
+/* Specific component adjustments */
+.rt-Skeleton {
+  &:where(.rt-SliderRoot) {
+    height: var(--slider-track-size);
+  }
+  &:where(.rt-TextFieldInput) + .rt-TextFieldChrome {
+    visibility: hidden;
+  }
+  .rt-TextFieldRoot:has(&) {
+    cursor: default !important;
+  }
+}
+
 @keyframes rt-pulse {
   from {
     background-color: var(--gray-a3);
--- a/packages/radix-ui-themes/src/components/skeleton.css
+++ b/packages/radix-ui-themes/src/components/skeleton.css
@@ -2,17 +2,19 @@
   border-radius: var(--radius-1);
 }
 
-.rt-Skeleton {
+.rt-Skeleton,
+.rt-Skeleton:where(.rt-TextFieldInput) + :where(.rt-TextFieldChrome) {
   animation: rt-skeleton-pulse 1000ms infinite alternate-reverse !important;
   background-image: none !important;
+  background-clip: border-box !important;
   border: none !important;
   box-shadow: none !important;
+  box-decoration-break: clone !important;
   color: transparent !important;
   outline: none !important;
   pointer-events: none !important;
   user-select: none !important;
   cursor: default !important;
-  box-decoration-break: clone !important;
 }
 
 /*
@@ -28,15 +30,10 @@
  *
  * Will NOT collapse Skeletons as the browsers use line-height on inline elements just for alignment.
  */
-.rt-Skeleton[data-inline-skeleton] {
+.rt-Skeleton:where([data-inline-skeleton]) {
   line-height: 0;
   /* Font metrics on Arial yield much nicer height for the text background bounding box */
   font-family: Arial, sans-serif !important;
-}
-
-.rt-Skeleton:autofill {
-  -webkit-text-fill-color: transparent !important;
-  box-shadow: inset 0 0 0 100px var(--gray-a3), inset 0 0 0 100px var(--color-background) !important;
 }
 
 :where(.rt-Skeleton:empty) {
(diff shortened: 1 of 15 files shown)


vladmoroz:
@benoitgrelard should we handle the text field issue on your screenshot? I saw it before in WorkOS UIs and understand that it happens because of our fancy prop forwarding, but can't come up with anything better than special-casing the text field in CSS.

benoitgrelard:
@vladmoroz I wasn't sure whether it was intentional or not, given that the WorkDS docs call out specifically usage with interactive components (calling out hover/focus states, etc):

CleanShot 2023-12-18 at 14 24 12@2x

What is the intended behaviour?

vladmoroz:
@benoitgrelard it's not intended, it got broken in WorkDS when we changed TextField to Radix Themes

benoitgrelard:
@vladmoroz I have done a pass on every single component https://github.com/radix-ui/themes/pull/188/commits/72d76eb9177b190bb5c8cc49d7cfd677230ad35c:

CleanShot 2023-12-18 at 23 12 37@2x

These are the adjustments I made:
- Add `inert` to `Skeleton`
- `Checkbox`, `RadioGroupItem`, `Slider` & `Switch`: Lift up `border-radius` so it is correct when wrapped in `Skeleton`
- Special CSS handling for:
  - `Slider`: Ensure the correct `height` is used when wrapped in `Skeleton`
  - `TextFieldInput`: Ensure correct visual and cursor used when wrapped in `Skeleton`

packages/radix-ui-themes/src/components/skeleton.css
```diff
@@ -0,0 +1,78 @@
...
+.rt-Skeleton[data-inline-skeleton] {
+  line-height: 0;
+  /* Font metrics on Arial yield much nicer height for the text background bounding box */
+  font-family: Arial, sans-serif !important;
+}
+
+.rt-Skeleton:autofill {
+  -webkit-text-fill-color: transparent !important;
+  box-shadow: inset 0 0 0 100px var(--gray-a3), inset 0 0 0 100px var(--color-background) !important;
+}
```
vladmoroz, line 40: - We should check if this is actually needed, this style was authored for the pre-RT Text Field
- If still needed, we can use the newer technique for removing the native autofill colors that we use in the Text Field itself

packages/radix-ui-themes/src/components/skeleton.css
```diff
@@ -0,0 +1,78 @@
...
+/* `visibility: hidden` doesn’t work for placeholders in Firefox */
+.rt-Skeleton::placeholder {
+  opacity: 0 !important;
+}
+
+/* Specific component adjustments */
+.rt-Skeleton {
+  &:where(.rt-SliderRoot) {
+    height: var(--slider-track-size);
+  }
```
vladmoroz, line 62: I think we shouldn't change dimensions so that Skeleton doesn’t affect layout. Checked the Slider and a slight mismatch between the Skeleton version and the real thing seems fine as a trade-off.

packages/radix-ui-themes/src/components/skeleton.css
```diff
@@ -0,0 +1,78 @@
...
+  &:where(.rt-TextFieldInput) + .rt-TextFieldChrome {
+    visibility: hidden;
+  }
+  .rt-TextFieldRoot:has(&) {
+    cursor: default !important;
+  }
+}
+
+@keyframes rt-pulse {
```
vladmoroz, line 71: Let's prefix this with component name, `rt-skeleton-pulse`?

packages/radix-ui-themes/src/components/skeleton.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const Tag = React.isValidElement(children) ? Slot : 'span';
+  return (
+    <Tag
+      ref={forwardedRef}
+      aria-hidden
+      className={classNames('rt-Skeleton', className, withMarginProps(marginProps))}
+      data-inline-skeleton={React.isValidElement(children) ? undefined : true}
+      tabIndex={-1}
+      // Workaround to use `inert` until https://github.com/facebook/react/pull/24730 is merged.
+      {...{ inert: true ? '' : undefined }}
```
vladmoroz, line 22: Curious if we still need `tabIndex`, `aria-hidden`, and `pointer-events` with inert?

packages/radix-ui-themes/src/components/skeleton.css
benoitgrelard: How would this state even trigger? You'd have to be auto-filling something and THEN it goes skeleton. Seems unlikely no?

packages/radix-ui-themes/src/components/skeleton.css
benoitgrelard: Fair, I did wonder about that when making the change. Right now basically it will become the size of the thumb.

packages/radix-ui-themes/src/components/skeleton.tsx
benoitgrelard: I feel like we should keep them in case `inert` isn't supported.

vladmoroz:
@benoitgrelard also let's bikeshed the API tomorrow, there's a couple things that have felt slightly annoying since we've been using it in WorkDS:
- Something like a `loading` prop idea has come up a few times, e.g. `<Skeleton loading={isLoading}>` to render it conditionally
- We should discuss whether Skeleton should support `width` and `height` props that'd take both arbitrary and space scale values. The way it is now feels like too much friction for building throwaway loading states.
  - Supporting both arbitrary and space scale values is something that has been on my mind for all layout components too; somewhere on a local branch I have a PoC of a much better approach than what we have for Grid. I can work on that as a next step once we agree on the API

packages/radix-ui-themes/src/components/skeleton.css
vladmoroz: Browsers will autofill forms you had submitted before on page load, check this out:

https://github.com/radix-ui/themes/assets/8441036/6760de0a-2f1a-44ac-be91-a62b5dc7bf3b

benoitgrelard:
> Something like a loading prop idea has come up a few times, e.g. <Skeleton loading={isLoading}> to render it conditionally. Otherwise the code gets easily littered with ternaries.

On that point, I'm not sure what's worse to be honest. I feel like most of the time we end up reproducing chunks of UIs in skeletons with an early return.

packages/radix-ui-themes/src/components/skeleton.css
benoitgrelard: I might need help testing this out, I can't get it to even auto-fill a whole email/password form in Chrome.

https://github.com/radix-ui/themes/assets/1539897/47b1583b-5bb1-4e1b-bef8-cfcdee5d4f99

packages/radix-ui-themes/src/components/skeleton.css
vladmoroz: Checking now, do you remember if it was deliberate that the Slider height matches thumb height? For layout reasons?

packages/radix-ui-themes/src/components/skeleton.css
vladmoroz: @benoitgrelard I think you need a form that you'd submit at least once to get that
Pushed a `test-skeleton` page to your branch with that and tested:

https://github.com/radix-ui/themes/assets/8441036/5806c1d2-7c10-428e-848f-cda320002d16

Looks like the workaround for autofill is still needed; and the current technique does have a cosmetic problem, you can see that the browser’s autofill kicks in half a second after you load the page and makes the animated elements feel out of sync.

Here’s what it looks like if we don't handle autofill:

https://github.com/radix-ui/themes/assets/8441036/31e98b0b-e6b5-443a-a905-e70500ce320b

vladmoroz:
> On that point, I'm not sure what's worse to be honest. I feel like most of the time we end up reproducing chunks of UIs in skeletons with an early return.

That’s partly a product of not having a smooth DX for the Skeleton. Also this would be even more relevant for UIs that are mostly static and have just a couple individual pieces that might need to load the data client-side; that's when you'd always do granular ternaries in the JSX.

packages/radix-ui-themes/src/components/skeleton.css
benoitgrelard: Yeah I think so.

vladmoroz (approved):

PR #193, opened 2023-12-20
Add `gridColumn` and `gridRow` props, rework Grid’s `columns` and `rows`

vladmoroz:
- Rework how `columns` and `rows` props are set on Grid to simplify the CSS
  - Pre-defined values for "1" to "9" are set with regular classes like "rt-gtc-1"
  - Arbitrary values are set through CSS custom properties using a combination of the classname and `style` attribute
    - The pre-defined values could work similarly (instead of having own classnames), however I think that there's some value in having a cleaner experience inspecting the DOM without the extra `style` attributes. For other properties we want to use this technique with, like width and height, it would also keep the existing classnames as is, maximising compatibility in fringe cases.
- Add `gridColumn`, `gridColumnStart`, `gridColumnEnd`, `gridRow`, `gridRowStart`, and `gridRowEnd` layout props.
  - @benoitgrelard any thoughts on whether to keep `gridColumnStart`, `gridColumnEnd`, `gridRowStart`, and `gridRowEnd`? Maybe too much on one hand, no harm to have the flexibility on the other hand.


packages/radix-ui-themes/src/helpers/props/prop-def.ts
```diff
@@ -18,14 +18,23 @@ type RegularPropDef<T> =
...
+};
+type PropDef<T = any> = RegularPropDef<T> | ResponsivePropDef<T> | ResponisveWidePropDef<T>;
 
 // prettier-ignore
 type GetPropDefType<Def> =
 	  Def extends BooleanPropDef ? (Def extends ResponsivePropDef ? Responsive<boolean> : boolean)
   : Def extends StringPropDef ? (Def extends ResponsivePropDef ? Responsive<string> : string)
   : Def extends StringOrNumberPropDef ? (Def extends ResponsivePropDef ? Responsive<string | number> : string | number)
   : Def extends ReactNodePropDef ? (Def extends ResponsivePropDef ? Responsive<React.ReactNode> : React.ReactNode)
+  : Def extends ResponisveWidePropDef<infer Type> ? (Def extends ResponisveWidePropDef<infer Type extends string> ? ResponsiveWide<Type> : Type)
```
vladmoroz, line 37: Not sure I nailed the types, especially here

packages/radix-ui-themes/src/helpers/props/layout.props.ts
```diff
@@ -136,10 +206,80 @@ function withLayoutProps(props: LayoutProps) {
...
+function getLayoutStyles(props: LayoutProps) {
+  const baseLayoutClassNamess = withLayoutProps(props);
+
+  const [gridColumnClassNames, gridColumnCustomProperties] = getResponsiveStyles({
+    allowAribtraryValues: true,
+    className: 'rt-r-gc',
+    customProperty: '--grid-column',
+    value: props.gridColumn,
+    values: layoutPropDefs.gridColumn.values,
+  });
```
vladmoroz, line 222: It's a bit lengthy to retrieve the classes and CSS vars like this, wondering if that’s fine or we should try to figure out a more succinct syntax for this stuff

packages/radix-ui-themes/src/helpers/breakpoints.ts
```diff
@@ -1,5 +1,9 @@
+import { hasOwnProperty } from './has-own-property';
+
 type Breakpoints = 'initial' | 'xs' | 'sm' | 'md' | 'lg' | 'xl';
 type Responsive<T> = T | Partial<Record<Breakpoints, T>>;
+type Wide<T extends string> = T | Omit<string, T>;
+type ResponsiveWide<T extends string> = Responsive<Wide<T>>;
```
vladmoroz, line 6: Do these names make sense? Any better ideas @benoitgrelard?

packages/radix-ui-themes/src/helpers/breakpoints.ts
```diff
@@ -67,11 +71,151 @@ function withBreakpoints(
...
+
+interface GetResponsiveClassNamesOptions {
+  className?: string;
+  value: ResponsiveWide<string> | Responsive<string> | undefined;
+  values?: string[] | readonly string[];
+  allowAribtraryValues?: boolean;
+  map?: (value: string | undefined) => string | undefined;
+}
+
+function getResponsiveClassNames({
```
vladmoroz, line 102: This function will ultimately replace `withBreakpoints`. It basically does the same thing, but accounts for creating corresponding classnames when an arbitrary value is passed.

It will need to work with boolean props first though.

packages/radix-ui-themes/src/helpers/props/layout.props.ts
```diff
@@ -1,8 +1,11 @@
-import { withBreakpoints } from '../breakpoints';
+import classNames from 'classnames';
+import { withBreakpoints, getResponsiveStyles } from '../breakpoints';
+import { styles } from '../styles';
 
 import type { PropDef, GetPropDefTypes } from './prop-def';
 
-const paddingValues = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'] as const;
+const positiveNumbers = ['1', '2', '3', '4', '5', '6', '7', '8', '9'] as const;
```
benoitgrelard, line 7: I feel like we shouldn't do this.

- For `padding`, `margin`, `width, `height`, those positive numbers for 0-9 refer to the steps in the space scale (finite)
- For `gridColumn` and `gridRow` those positive numbers refer to a number of equal columns/rows (infinite really, could be 22 or 78, kinda arbitrary)
- For `gridColumnStart`, `gridColumnEnd`, `gridRowStart`, `gridRowEnd` those positive numbers would refer to the column/grid lines/names etc (so 9 wouldn't make sense anyway if it's 3 column grid, also -1 is a possible value)

Not sure how we want to handle the grey area with all this, but I feel like we shouldn't conflate in code like that, especially with pure space scale stuff.

packages/radix-ui-themes/src/helpers/props/prop-def.ts
```diff
@@ -18,14 +18,23 @@ type RegularPropDef<T> =
   | ReactNodePropDef
   | EnumPropDef<T>;
 type ResponsivePropDef<T = any> = RegularPropDef<T> & { responsive: true };
-type PropDef<T = any> = RegularPropDef<T> | ResponsivePropDef<T>;
+type ResponisveWidePropDef<T = any> = {
```
benoitgrelard, line 21:
```suggestion
type ResponsiveWidePropDef<T = any> = {
```

packages/radix-ui-themes/src/helpers/styles.ts
```diff
@@ -0,0 +1,14 @@
+type InlineStyle = React.CSSProperties | Record<string, string | number | undefined> | undefined;
+
+// Merges CSS styles like `classNames` merges CSS classes
+export function styles(...styles: Array<InlineStyle>): InlineStyle {
```
benoitgrelard, line 4: I get the analogy with the `classNames` package, but it took me some time to realise what this was until I saw the util and the comment.

I wonder if we should just name it `mergeStyles`?

packages/radix-ui-themes/src/helpers/breakpoints.ts
benoitgrelard: Not really no, what are we trying to convey here?
- an enum with a list of options
- but also the rest of all strings
- does it have to be responsive?

packages/radix-ui-themes/src/styles/utilities/grid-column.css
```diff
@@ -0,0 +1,63 @@
...
+  }
+  .rt-r-gc-7 {
+    grid-column: 7;
+  }
+  .rt-r-gc-8 {
+    grid-column: 8;
+  }
+  .rt-r-gc-9 {
+    grid-column: 9;
+  }
```
benoitgrelard, line 31: I'm probably not understanding all of this correctly, but if we support arbitrary values, why do we even need to generate all of this CSS?

Wouldn't it work anyway with via the `custom properties`?

packages/radix-ui-themes/src/components/box.tsx
```diff
@@ -24,7 +25,17 @@ interface BoxProps
...
+
+  const {
+    className,
+    asChild,
+    display = boxPropDefs.display.default,
+    style,
+    ...boxProps
+  } = layoutRest;
+
+  const [layoutClassNames, layoutCustomProperties] = getLayoutStyles(layoutProps);
```
benoitgrelard, line 37: Wouldn't all the other layout components like `Flex`, ``Container`, `Section` also need all those changes?

CleanShot 2023-12-20 at 12 30 51@2x

packages/radix-ui-themes/src/helpers/breakpoints.ts
```diff
@@ -67,11 +71,151 @@ function withBreakpoints(
   return classes.join(' ');
 }
 
+function getResponsiveStyles({
+  allowAribtraryValues,
```
benoitgrelard, line 75:
```suggestion
  allowArbitraryValues,
```

benoitgrelard (commented):
This is pretty dense stuff, I've added a few comments for now as I have some questions.

packages/radix-ui-themes/src/helpers/props/layout.props.ts
benoitgrelard: Equally, for `gridColumns` and `gridRows` we have 0-9 but that's kinda arbitrary as it's a number of equal columns/rows (not finite and unrelated to space scale)

packages/radix-ui-themes/src/helpers/breakpoints.ts
vladmoroz: > - an enum with a list of options
> - but also the rest of all strings

Yep!

> does it have to be responsive?

Doe not **have** to be, it theory we could have a prop that’s just `Wide` but not `ResponsiveWide`. However, I'm not seeing a case for one yet. All the props that will support wider values happen to be responsive.

packages/radix-ui-themes/src/components/box.tsx
vladmoroz: Yep, adding

vladmoroz:
Handled your feedback @benoitgrelard

For posterity, some decisions we discussed:
- When a prop supports arbitrary values, its types will suggest only theme values or values with special meaning (e.g. `<Box m="2" />` for margin on the theme scale or `<Grid columns="2" />` for two equal grid columns).
- When a prop supports arbitrary values, but doesn’t accept any theme values or values with special meanings, the prop types will suggest the most common keywords or their aliases (e.g. `<Flex align="start" />`).
- We’ll use a combination of pre-defined CSS classes and on-the-fly CSS custom properties to keep the DOM tidy for better DX/debugging experience. Arbitrary values are rather a feature on top the pre-defined utilities we wire up the props to. The analogy here is how now, most of the layout works now with CSS utilities under the hood, but the arbitrary values have to be defined in the `style` attribute. With the new implementation, the arbitrary values would also end up in the `style` attribute as well, but with the extra flexibility of the breakpoints object.
- We’ll add JSDoc comments documentation to the props on layout components
  - I’ll do this separately when we get through supporting the new system for all layout props

benoitgrelard (approved):

PR #200, opened 2023-12-22
Expose more props to docs

benoitgrelard:


packages/radix-ui-themes/src/components/progress.tsx
```diff
@@ -13,9 +13,7 @@ type ProgressOwnProps = GetPropDefTypes<typeof progressPropDefs>;
 interface ProgressProps
   extends Omit<PropsWithoutRefOrColor<typeof ProgressPrimitive.Root>, 'children'>,
     MarginProps,
-    ProgressOwnProps {
-  duration?: string;
-}
```
vladmoroz, line 18: Should we do specify the duration type so that `${number}s` is enforced?

vladmoroz (approved):

packages/radix-ui-themes/src/components/progress.tsx
vladmoroz: We can keep it as 'string' in the docs

benoitgrelard: you mean, override the type in the component?

vladmoroz: Yeah potentially

PR #203, opened 2023-12-22
Add loading to `Spinner`, `IconButton` and `Button`

benoitgrelard:
https://github.com/radix-ui/themes/assets/1539897/4da34d85-3086-4366-875d-721a399c0037

### DIFF
--- a/apps/playground/components/loading-buttons.tsx
+++ b/apps/playground/components/loading-buttons.tsx
@@ -2,7 +2,17 @@
 
 import * as React from 'react';
 import { ArrowRightIcon, StarIcon } from '@radix-ui/react-icons';
-import { Button, Flex, IconButton, Spinner } from '@radix-ui/themes';
+import {
+  Button,
+  Em,
+  Flex,
+  IconButton,
+  Spinner,
+  Text,
+  buttonPropDefs,
+  mapButtonSizeToSpinnerSize,
+  mapResponsiveProp,
+} from '@radix-ui/themes';
 
 export function LoadingButtons() {
   const [loading1, onClick1] = useLoading();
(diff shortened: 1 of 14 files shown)


packages/radix-ui-themes/src/components/base-button.tsx
```diff
@@ -18,19 +21,22 @@ const BaseButton = React.forwardRef<BaseButtonElement, BaseButtonProps>((props,
...
     className,
+    children,
     asChild = false,
     size = baseButtonPropDefs.size.default,
     variant = baseButtonPropDefs.variant.default,
     color = baseButtonPropDefs.color.default,
     highContrast = baseButtonPropDefs.highContrast.default,
     radius = baseButtonPropDefs.radius.default,
+    loading = baseButtonPropDefs.loading.default,
+    disabled = loading,
```
vladmoroz, line 32: Whaaat this is legal?

packages/radix-ui-themes/src/components/base-button.tsx
```diff
@@ -41,10 +47,38 @@ const BaseButton = React.forwardRef<BaseButtonElement, BaseButtonProps>((props,
...
+           * However, `display: contents` removes the content from the accessibility tree in some browsers,
+           * so we force remove it with `aria-hidden` and re-add it in the tree with `VisuallyHidden`
+           */}
+          <span style={{ display: 'contents', visibility: 'hidden' }} aria-hidden>
+            {children}
+          </span>
+          <VisuallyHidden>{children}</VisuallyHidden>
+
+          <Flex asChild align="center" justify="center" position="absolute" inset="0">
+            <span>
```
vladmoroz, line 73: Is this not to avoid putting a div into buttons?

apps/playground/components/loading-buttons.tsx
vladmoroz, line 1: Can we add a demo for each button size? We might want to tweak the default spinner size for when the `loading` prop is used

vladmoroz (commented):
The spinner within disabled buttons looks quite light, should we adjust the color it uses in this case?

packages/radix-ui-themes/src/components/base-button.tsx
benoitgrelard: haha yep

packages/radix-ui-themes/src/components/base-button.tsx
benoitgrelard: I'm not sure if it's strictly needed, but my legacy brains takes over 😀 
It kinda just feels cleaner with `span` when looking at the DOM too.

benoitgrelard:
> The spinner within disabled buttons looks quite light, should we adjust the color it uses in this case?

Yeah I was thinking the same, the text color is `gray-a8` for any disabled button right now, we could have a special rule in `spinner` like:

```css
:where(.rt-BaseButton[data-disabled]) & {
  color: ? // something else instead of currentColor
}
```

Here it is with `gray-a11` instead:

https://github.com/radix-ui/themes/assets/1539897/80b22af7-e9c3-48cd-8895-679bb27e3e44

packages/radix-ui-themes/src/components/base-button.tsx
vladmoroz: Yeah I think it's one of the reasons we should have as="span"

vladmoroz: Technically spans in buttons are considered wrong afair

benoitgrelard:
Ready to review again @vladmoroz

vladmoroz (approved):

PR #221, opened 2024-01-08
Rename flex props

vladmoroz:
Deprecate `shrink` and `grow` props in favour of `flexShrink` and `flexGrow`. The older `shrink` and `grow` props will be removed in the next major release.


vladmoroz:
Closing in favor of #235

PR #226, opened 2024-01-10
Implement arbitrary layout props

vladmoroz:
**Changelog**

- General
  - Add `flexBasis`, `minWidth`, `maxWidth`, `minHeight`, `maxHeight` props to the layout components
  - Rework all layout props to allow arbitrary CSS values, including when used with the responsive object syntax. Props that support arbitrary values include `width`, `minWidth`, `maxWidth`, `height`, `minHeight`, `maxHeight`, `p`, `px`, `py`, `pt`, `pr`, `pb`, `pl`, `inset`, `top`, `right`, `bottom`, `left`, `flexBasis`, `flexShrink`, `flexGrow`, `gridColumn`, `gridColumnStart`, `gridColumnEnd`, `gridRow`, `gridRowStart`, `gridRowEnd`.
  - Update the type signature of the layout props so that code editor suggestions use just space scale values when possible. CSS keywords and other values such as `"auto"` or `"100vw"` are still available as a manual string values.
- `Table`
  - Add new `layout` prop to control the `table-layout` style property
  - Align `width` prop type signature and implementation on the `TableCell` part with the reworked `width` prop on the layout components
  - Add `minWidth` and `maxWidth` props to the `TableCell` part


packages/radix-ui-themes/src/components/inset.tsx
```diff
@@ -32,19 +32,23 @@ const Inset = React.forwardRef<InsetElement, InsetProps>((props, forwardedRef) =
...
         withMarginProps(marginProps)
       )}
     />
   );
 });
 Inset.displayName = 'Inset';
 
+const paddingMap = {
+  current: 'inset',
+};
```
vladmoroz, line 51: I changed `rt-r-p-current` classname to `rt-r-p-inset` so the name of the class reflects the implementation (e.g. in case we have another component unrelated to Inset with "current" padding value)

packages/radix-ui-themes/src/helpers/breakpoints.ts
```diff
@@ -154,7 +154,7 @@ function getBaseClassName(
 
 interface GetResponsiveCustomPropertiesOptions {
   value: Responsive<StringOrValue<string>> | Responsive<string> | undefined;
-  customProperty: `--${string}`;
+  customProperty: `--${string}` | `--${string}`[];
```
vladmoroz, line 157: This is needed for props like `px` which need to set both `padding-left` and `padding-right`. The type signature is a bit weird because I needed something that would work with as few changes as possible, but I want to rework it in the next PR that would refactor classnames/styles to be driven by prop defs.

vladmoroz:
Closing in favor of #235

Issue #205, opened 2023-12-23
[Feature request] Allow the `color` props on `Tabs`


pradel:
I want to be able to set my Tabs as `gray` and have the underline render the right color for example.

vladmoroz:
Hey, that’s oversight on our part, we forgot to expose a prop for that. We’ll release a fix after the holidays, for now you can set a `data-accent-color="gray"` prop on the Tab List part as a workaround.

kendallstrautman:
@pradel This was fixed in https://github.com/radix-ui/themes/pull/228 and will be included in the next release!

PR #228, opened 2024-01-10
tabs: Pass color prop

kendallstrautman:
## Description

Support the `color` prop to control the color of the active tab trigger. 

Screenshot 2024-01-10 at 9 41 50 AM

## Testing steps

Visit the [sink page ](https://themes-playground-git-ksadd-tabs-color-prop-modulz.vercel.app/sink#Tabs)to verify the colors are properly set in the examples and the default is the global accent color.

## Relates issues / PRs

Addresses #205

### DIFF
--- a/packages/radix-ui-themes/src/components/tabs.css
+++ b/packages/radix-ui-themes/src/components/tabs.css
@@ -106,7 +106,7 @@
   --tabs-list-active-accent: var(--accent-10);
 
   &.rt-high-contrast {
-    --tabs-list-active-accent: var(--accent-11);
+    --tabs-list-active-accent: var(--accent-12);
   }
 }
 
--- a/packages/radix-ui-themes/src/components/tabs.css
+++ b/packages/radix-ui-themes/src/components/tabs.css
@@ -137,7 +137,7 @@
     background-color: var(--accent-10);
   }
 
-  .rt-TabsList.rt-high-contrast & {
+  :where(.rt-TabsList.rt-high-contrast) & {
     &:where([data-state='active'])::before {
       background-color: var(--accent-12);
     }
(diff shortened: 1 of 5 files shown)


vladmoroz (commented):
- Let’s put the `color` prop on the List part—otherwise, when on the Root, it provides a new accent color also for the children of Tabs Content:
  
  image

  This property of `data-accent-color` is by design and is desirable in most cases otherwise.
- Let’s add `highContrast` prop so that you can set a darker accent color. I believe all or almost all components with `color` prop have a `highContrast` option
- Could you also note the change in `changelog.md` under 2.1.0?

kendallstrautman:
@vladmoroz ty for taking a look. Will update the color to be set on the list and also update the changelog / docs. 

> Let’s add highContrast prop

Sure thing! Would `accent-11` work for high contrast mode? I'm not sure how far to step for the colors with contrast.

vladmoroz (approved):

packages/radix-ui-themes/src/components/tabs.css
```diff
@@ -134,7 +139,7 @@
     bottom: 0;
     left: 0;
     right: 0;
-    background-color: var(--accent-10);
+    background-color: var(--tabs-list-active-accent);
```
vladmoroz, line 142: Nit:

We usually opt to use CSS variables almost as the last resort to handle multiple style dependencies related to the same value across different parts of the component.

Otherwise, we’d rather trade off larger selector complexity for being able to read the values right where they are applied, like this:

```suggestion
    background-color: var(--accent-10);
    
    :where(.rt-TabsTrigger.rt-high-contrast) & {
      background-color: var(--accent-12);
    }
  }
```

kendallstrautman: Good to know, thank you!

PR #231, opened 2024-01-12
[b] Add DataList component with `direction`

kendallstrautman:
🚧 WIP

## Description

[See testing page](https://themes-playground-git-ksadds-data-list-with-direction-modulz.vercel.app/test-data-list)

Another option related to #230 that uses `direction` to set the layout and retains `width` being handled by the labels. One option here would be to take a `width` property at the root component and set it for the labels. I do prefer to composability of this interface over the column approach even if it requires slightly more repetition. 

```jsx
// Use the flex properties `column` or `row` to denote the kv pair layout
<DataListRoot mt="2" direction="column">
  <DataListItem>
    <DataListLabel>Name</DataListLabel>
    <DataListData>Jane Roe</DataListData>
  </DataListItem>
  <DataListItem>
    <DataListLabel>Email</DataListLabel>
    <DataListData>janeroe@foo-corp.com</DataListData>
  </DataListItem>
  <DataListItem>
    <DataListLabel>Organization</DataListLabel>
    <DataListData>
      <Link href="https://foo-corp.com">Foo Corp</Link>
    </DataListData>
  </DataListItem>
</DataListRoot>
```

### DIFF
--- a/apps/playground/app/test-data-list/page.tsx
+++ b/apps/playground/app/test-data-list/page.tsx
@@ -1,0 +1,68 @@
+import Link from 'next/link';
+import {
+  DataListRoot,
+  DataListData,
+  DataListItem,
+  DataListLabel,
+  Theme,
+  Container,
+  Text,
+} from '@radix-ui/themes';
+
+export default function DataListPage() {
+  return (
+    <html>
+      <body>
+        <Theme>
+          <Container m="6">
+            <Text weight="medium">Default</Text>
+            <DataListRoot mt="2">
+              <DataListItem>
+                <DataListLabel>Obi wan</DataListLabel>
+                <DataListData>Jedi Master</DataListData>
+              </DataListItem>
+              <DataListItem>
+                <DataListLabel>Anakin</DataListLabel>
+                <DataListData>Padewan</DataListData>
+              </DataListItem>
+            </DataListRoot>
+          </Container>
+          <Container m="6">
+            <Text weight="medium">Vertical Layout</Text>
+            <DataListRoot layout="vertical" mt="2">
+              <DataListItem>
+                <DataListLabel>Obi wan</DataListLabel>
+                <DataListData>Jedi Master</DataListData>
+              </DataListItem>
+              <DataListItem>
+                <DataListLabel>Anakin</DataListLabel>
+                <DataListData>Padewan</DataListData>
+              </DataListItem>
+            </DataListRoot>
+          </Container>
+          <Container m="6">
+            <Text weight="medium">Varied data</Text>
+            <DataListRoot mt="2">
+              <DataListItem>
+                <DataListLabel>Name</DataListLabel>
+                <DataListData>Jane Roe</DataListData>
+              </DataListItem>
+
+              <DataListItem>
+                <DataListLabel>Email</DataListLabel>
+                <DataListData>janeroe@foo-corp.com</DataListData>
+              </DataListItem>
+
+              <DataListItem>
+                <DataListLabel>Organization</DataListLabel>
+                <DataListData>
+                  <Link href="https://foo-corp.com">Foo Corp</Link>
+                </DataListData>
+              </DataListItem>
+            </DataListRoot>
+          </Container>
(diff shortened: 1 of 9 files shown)


vladmoroz:
Have synced today with Benoit, testing both [a] and [b] versions, and this direction does seem more promising, especially when factoring in that the best TypeScript experience achievable with [a] feels like the weirdest thing ever—I saw your todo there on the prop defs, but we bypassed them and tried to design an ideal type for that prop right on the component interface itself, and it didn't feel good the way VSC suggestions worked.

So anyway, let's settle on `direction` for now. Individual `width` might be fine, but the other thing with it is the behaviour it has in the table layout we have:
- Effectively, width acts as min-width in most table layouts
- There’s no responsive prop for max-width in the WorkDS component, but even if there were, it doesn't work as expected in table layouts

Which brings me to the question—do you think we can pull off an equivalent component with a `display: grid` layout? (We can ignore the issue with the `trim` technique and address it separately). That would allow us to implement a control for label width that works as expected with `width`, `min-width`, and `max-width` properties, added as props in #226

vladmoroz:
Checked [c] – I'll keep all feedback here so it's easier to track of:
- Using a native gap does feel like it simplifies everything a ton. Not only for ourselves, but also for the developer using this component and inspecting it, which is super valuable imo.
- Having the value width potentially out of sync is a huge bummer. The ideal experience here is that you wouldn't have to control the default width most of the time, and in case you do have a long label somewhere, it all gets aligned automatically. Before we did this component we saw this situation (label widths out of sync) quite a few times in our product UIs.
- Grid layout would be ideal, but as you said then the potential tradeoff would be that we need to get rid of the item part.

This leads me to the following idea: we can keep the item part, but have it use `display: subgrid`.
- The [support](https://caniuse.com/?search=subgrid) is decent enough already
  - It is worst supported on mobile browsers where the UIs would normally use a vertical layout anyway,
  - The items can fall back to `display: grid`, with the tradeoff of having the value width potentially out of sync only in the older browsers.
    - Grid-related styles can be inherited from the parent
- The API could be what you have here on this PR, but with display: grid/subgrid implemention, plus `width`, `minWidth`, `maxWidth` values on the label parts
  - We can still consider `columns` but I am not sure how to tie it in with `direction` (or to replace it altogether).
- Trim can be re-created in a custom way for this component, using negative margin on the first and last item that would use the same calculations as `.rt-r-lt` classes
- Altogether, in modern browsers we'd get good DX when debugging and inspecting the layout, and width styles on the labels that work as expected

apps/playground/app/test-data-list/page.tsx
```diff
@@ -0,0 +1,242 @@
...
+
+                <DataListItem align="center">
+                  <DataListLabel>Authentication methods</DataListLabel>
+                  <DataListData>
+                    <Flex gap="2">
+                      <StarFilledIcon />
+                      <StarFilledIcon />
+                    </Flex>
+                  </DataListData>
+                </DataListItem>
```
vladmoroz, line 154: Looks like `align="center"` doesn't quite work in the case when elements in `DataListData` have smaller height than the line-height of the label.

When we implement it for real, let's make sure it works. Also we need a similar visual example for `align="center"`, but when `DataListData` elements are larger in height (which was the original test here).

apps/playground/app/test-data-list/page.tsx
```diff
@@ -0,0 +1,242 @@
...
+                With nested flex
+              </Heading>
+              <DataListRoot>
+                <DataListItem>
+                  <DataListLabel width={80}>Appearance</DataListLabel>
+                  <DataListData>
+                    <Flex align="center" gap="1">
+                      <IconButton size="1">
+                        <InfoCircledIcon />
+                      </IconButton>
```
vladmoroz, line 177: This should check for whether baseline alignment "just works" when data is a Flex with text that has a small icon or a small button aside.

Let's use `variant="ghost"` on the button the plain icon, otherwise a larger button pushes this element down and there is no expectation that this should work.

PR #237, opened 2024-01-16
Add DataList component

kendallstrautman:
## Description

Adds a component for displaying a group of key/value pairs. The `orientation` prop controls the orientation between the key value pair, whether horizontally or vertically laid out. 

Screenshot 2024-01-17 at 4 16 07 PM

## Testing steps

See [sink page](https://themes-playground-git-ksadds-data-list-grid-subgrid-modulz.vercel.app/sink#Data%20List) for examples and also the example in [explore components](https://themes-playground-git-ksadds-data-list-grid-subgrid-modulz.vercel.app/explore-components#data-list).

## Todo

- [x] Add changelog note
- [x] Cleanup example code generation in sink and explore-components

## Relates issues / PRs

#230 
#231 
#233


apps/playground/app/test-data-list/page.tsx
```diff
@@ -0,0 +1,228 @@
...
+                  </DataListData>
+                </DataListItem>
+              </DataListRoot>
+            </Box>
+
+            <Box mb="6">
+              <Heading mb="5" size="3">
+                With custom `columns`
+              </Heading>
+              <DataListRoot columns="100px 1fr">
```
kendallstrautman, line 212: Since we're using subgrid, really long labels naturally will grow the width of the label column. I think the use of `columns` will be an edge case since the grid / subgrid interaction feels really nice / responsive as a default.

packages/radix-ui-themes/src/components/data-list.css
```diff
@@ -0,0 +1,191 @@
...
+  .rt-DataListRoot.rt-r-gap-y-9 {
+    column-gap: var(--space-9);
+  }
+}
+
+.rt-DataListRoot {
+  margin: 0;
+  display: grid;
+  /** Value  set via props in component **/
+  grid-template-columns: var(--data-list-columns, minmax(200px, max-content) 1fr);
```
kendallstrautman, line 110: I chose this default `minmax(200px, max-content) 1fr)` so the label column width will extend automatically even with really long labels. See this example below that doesn't require a custom label item width. 
Screenshot 2024-01-16 at 3 55 37 PM
Screenshot 2024-01-16 at 3 55 58 PM

packages/radix-ui-themes/src/components/data-list.css
```diff
@@ -0,0 +1,191 @@
...
+    align-items: start;
+  }
+
+  .rt-DataListItem.rt-r-vaf-middle {
+    align-items: center;
+  }
+
+  .rt-DataListItem.rt-r-vaf-end {
+    align-items: end;
+  }
```
kendallstrautman, line 143: @vladmoroz [re:  your comment here](https://github.com/radix-ui/themes/pull/231#discussion_r1453790691), had to rework `align` to work with flex.

apps/playground/app/test-data-list/page.tsx
```diff
@@ -0,0 +1,228 @@
...
+                </DataListItem>
+
+                <DataListItem align="center">
+                  <DataListLabel>Authentication methods</DataListLabel>
+                  <DataListData>
+                    <Flex gap="2" align="center" height="100%">
+                      <StarIcon />
+                      <MagicWandIcon />
+                      <StarFilledIcon />
+                    </Flex>
```
kendallstrautman, line 103: @vladmoroz https://github.com/radix-ui/themes/pull/231#discussion_r1453793293, we need to set the align on the `Flex` parent for this case which I think isn't odd given how flex works. Lmk your thoughts.

apps/playground/app/test-data-list/page.tsx
kendallstrautman: After playing with `grid/subgrid` I didn't feel that setting `width` on the label children made sense anymore and passing `columns` felt more appropriate. Of course this would only apply to the `row` direction, so we'd need to document that. Open to feedback on this!

packages/radix-ui-themes/src/components/data-list.props.ts
```diff
@@ -0,0 +1,79 @@
...
+    className: 'rt-r-vaf',
+    values: alignValues,
+    default: 'baseline',
+    responsive: true,
+  },
+} satisfies {
+  align?: PropDef<(typeof alignValues)[number]>;
+};
+
+export const dataListLabelPropDefs = {
```
kendallstrautman, line 59: The normal `width` prop defs that use 1-9 don't really work in this usecase since the spacing needed is much larger. Lmk if there's a preferred way to do this.

packages/radix-ui-themes/src/components/data-list.props.ts
```diff
@@ -0,0 +1,79 @@
+import { PropDef, trimProp, widthPropDefs } from '../helpers';
+
+const orientationValues = ['horizontal', 'vertical'] as const;
+const gapValues = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'] as const;
+const alignValues = ['start', 'center', 'end', 'baseline'] as const;
+const sizes = ['1', '2', '3'] as const;
```
kendallstrautman, line 6: We need to decide visually how we want these three sizes to look. Right now they map to the corresponding `Text` sizes.

apps/playground/app/test-data-list/page.tsx
```diff
@@ -0,0 +1,232 @@
+import Link from 'next/link';
```
kendallstrautman, line 1: 💡 I could delete this or move some examples into 'sink'

packages/radix-ui-themes/src/components/data-list.props.ts
```diff
@@ -0,0 +1,83 @@
...
+  size: {
+    type: 'enum',
+    className: 'rt-r-size',
+    values: sizes,
+    default: '2',
+    responsive: true,
+  },
+  trim: {
+    ...trimProp,
+    className: 'rt-r-dl-lt', // Custom trim styles due to grid layout
```
kendallstrautman, line 40: I used `dl` for data list, not sure exactly the best name to use here.

packages/radix-ui-themes/src/components/data-list.props.ts
kendallstrautman: Screenshot 2024-01-17 at 5 34 14 PM

packages/radix-ui-themes/src/components/data-list.css
```diff
@@ -0,0 +1,98 @@
...
+  .rt-r-o-horizontal .rt-DataListItem {
+    display: grid;
+    /* Use subgrid so all the label columns remain aligned */
+    grid-template-columns: subgrid;
+    padding: 0;
+    gap: inherit;
+  }
+
+  /* Align properties */
```
kendallstrautman, line 31:
```suggestion
  /* Align */
```

packages/radix-ui-themes/src/components/data-list.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+const DataListData = React.forwardRef<HTMLElement, React.ComponentPropsWithRef<'dd'>>(
+  ({ children, className, ...props }, forwardedRef) => (
+    <dd ref={forwardedRef} className={classNames(className, 'rt-DataListData')} {...props}>
+      {children}
+    </dd>
+  )
+);
+
+DataListData.displayName = 'DataListData';
```
benoitgrelard, line 73: Can we follow this format for each part?

```suggestion
type DataListDataElement = React.ElementRef<'dd'>;
interface DataListDataProps extends React.ComponentPropsWithoutRef<'dd'> {}
const DataListData = React.forwardRef<DataListDataElement, DataListDataProps>(
  ({ className, ...props }, forwardedRef) => (
    <dd {...props} ref={forwardedRef} className={classNames(className, 'rt-DataListData')} />
  )
);
DataListData.displayName = 'DataListData';
```

Specifically:
- starting with a type for the element and an interface for the props and then referencing those in forwardRef
- only pulling out `children` if necessary
- adding merged props after the spread of `{...props}` so it's clear what's been merged already vs. what will override or not the props
- no space before the display name

packages/radix-ui-themes/src/components/data-list.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export { DataListRoot, DataListItem, DataListLabel, DataListData };
```
benoitgrelard, line 75: We expose both compound and parts usually as well as the props:

```suggestion
const DataList = Object.assign(
  {},
  {
    Root: DataListRoot,
    Item: DataListItem,
    Label: DataListLabel,
    Data: DataListData,
  }
);

export { DataList, DataListRoot, DataListItem, DataListLabel, DataListData };
export type { DataListRootProps, DataListItemProps, DataListLabelProps, DataListDataProps };
```

packages/radix-ui-themes/src/components/index.ts
```diff
@@ -143,6 +143,8 @@ export { Callout, CalloutRoot, CalloutIcon, CalloutText } from './callout';
 export * from './callout.props';
 export { Card } from './card';
 export * from './card.props';
+export { DataListRoot, DataListData, DataListItem, DataListLabel } from './data-list';
```
benoitgrelard, line 146:
```suggestion
export { DataList, DataListRoot, DataListData, DataListItem, DataListLabel } from './data-list';
```

packages/radix-ui-themes/src/components/data-list.tsx
```diff
@@ -0,0 +1,75 @@
...
+  extends React.ComponentPropsWithoutRef<'dl'>,
+    MarginProps,
+    DataListRootOwnProps {}
+const DataListRoot = React.forwardRef<HTMLDListElement, DataListRootProps>(
+  (props, forwardedRef) => {
+    const { className, ...dataListProps } = extractProps(props, dataListPropDefs, marginPropDefs);
+    return (
+      <dl
+        ref={forwardedRef}
+        className={classNames(className, 'rt-DataListRoot', 'rt-Text')}
```
benoitgrelard, line 17: Typically we try not to rely just on classes like this (`rt-Text`).
Is there a reason here?

packages/radix-ui-themes/src/components/data-list.tsx
```diff
@@ -0,0 +1,75 @@
...
+      ref={forwardedRef}
+      className={classNames(className, 'rt-DataListLabel')}
+      {...labelProps}
+      style={
+        {
+          '--data-list-label-width': width,
+          '--data-list-label-min-width': minWidth,
+          '--data-list-label-max-width': maxWidth,
+          ...labelProps.style,
+        } as React.CSSProperties
```
benoitgrelard, line 57: Depending on my other point on custom properties, we may not need this.
If we do though, we should use our `mergeStyles` utility.

packages/radix-ui-themes/src/components/data-list.css
```diff
@@ -0,0 +1,98 @@
+/* Root -------------------------------*/
```
benoitgrelard, line 1: I don't think we need these comments, we don't typically have it in other files

packages/radix-ui-themes/src/components/data-list.css
```diff
@@ -0,0 +1,98 @@
...
+.rt-DataListRoot {
+  margin: 0;
```
benoitgrelard, line 3: Is this for reset purpose?
If so, should we add the things we're resetting to `reset.css` and add the `rt-reset` class instead?

packages/radix-ui-themes/src/components/data-list.css
```diff
@@ -0,0 +1,98 @@
...
+
+/* Item -------------------------------*/
+.rt-DataListItem {
+  position: relative;
+  box-sizing: border-box;
+  grid-column: span 2;
+}
+
+@breakpoints {
+  /* Orientation */
```
benoitgrelard, line 17: These however I would keep and actually use big comment blocks like we have already for SIZES/VARIANTS, etc

This way I would group by prop, rather than group by part for those styles.
If you look at other components, you'll see what I mean, basically you'd find all the code necessary for a specific prop under the same block.

packages/radix-ui-themes/src/components/data-list.css
benoitgrelard, line 1: Generally for the CSS we need to pay attention to CSS specificity.
We have the `lint:css` task which should help with that, but if you look at other CSS files, you'll see that we use `:where` extensively to keep specificity low.

benoitgrelard (commented):
Hey @kendallstrautman! This is getting close!
I've done a good first pass review on this.
Let me know if you have any questions!

apps/playground/app/explore-components/page.tsx
```diff
@@ -180,6 +185,7 @@ import {
   themeAccentColorsOrdered,
   //
   ThemePanel,
+  dataListPropDefs,
```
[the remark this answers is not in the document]
kendallstrautman: Thanks! Missed this auto-import.

packages/radix-ui-themes/src/components/data-list.css
kendallstrautman: In looking further I don't think this is needed, removing.

packages/radix-ui-themes/src/components/data-list.tsx
kendallstrautman: Good to know! Thank you.

packages/radix-ui-themes/src/components/data-list.tsx
kendallstrautman: I will refactor to use `Text` again!

packages/radix-ui-themes/src/components/data-list.tsx
```diff
@@ -0,0 +1,72 @@
+import classNames from 'classnames';
+import * as React from 'react';
+import { MarginProps, GetPropDefTypes, extractProps, marginPropDefs } from '../helpers';
+import { dataListPropDefs, dataListItemPropDefs, dataListLabelPropDefs } from './data-list.props';
```
benoitgrelard, line 4: nit: can we split out type and code imports as we do in other files?

packages/radix-ui-themes/src/components/data-list.tsx
```diff
@@ -0,0 +1,72 @@
...
+        className={classNames(className, 'rt-DataListRoot', 'rt-Text')}
+        {...dataListProps}
+      />
+    );
+  }
+);
+DataListRoot.displayName = 'DataListRootGrid';
+
+interface DataListItemProps
+  extends React.ComponentPropsWithRef<'div'>,
```
benoitgrelard, line 26: These should always use `ComponentPropsWithoutRef` when using `React.forwardRef` because then the typed `forwardRef` is the one adding the `ref` type./

packages/radix-ui-themes/src/components/data-list.css
kendallstrautman: Thanks the lint command was super helpful! 😄

packages/radix-ui-themes/src/components/data-list.tsx
kendallstrautman: So I refactored the `width`, `minWidth` etc.... to use custom properties and also support a range of enum sizes. Lmk what you think, I basically used the base [`width` as a starting point](https://github.com/radix-ui/themes/blob/main/packages/radix-ui-themes/src/styles/utilities/width.css).

apps/playground/app/sink/page.tsx
```diff
@@ -4725,6 +4728,256 @@ export default function Sink() {
...
+                          <DataListItem>
+                            <DataListLabel>Lorem ipsum dolor</DataListLabel>
+                            <DataListData>
+                              <Link href="https://workos.com">WorkOS</Link>
+                            </DataListData>
+                          </DataListItem>
+                        </DataListRoot>
+                      </Box>
+
+                      <Box mb="6">
```
kendallstrautman, line 4926: TODO: clean this up, make more generative.

packages/radix-ui-themes/src/components/data-list.css
```diff
@@ -0,0 +1,210 @@
...
+    }
+    &:where(.rt-r-dl-width-1) {
+      width: calc(var(--space-1) * var(--dl-width-scale));
+    }
+    &:where(.rt-r-dl-width-2) {
+      width: calc(var(--space-2) * var(--dl-width-scale));
+    }
+    &:where(.rt-r-dl-width-3) {
+      width: calc(var(--space-3) * var(--dl-width-scale));
+    }
```
kendallstrautman, line 126: Using the base `space` values and a scale multiplier. Totally open to another approach but wanted the widths to stay connected to the base space system. If we like this we can also add more sizes if we think that would be helpful / meaningful. 

[See related comment.](https://github.com/radix-ui/themes/pull/237#discussion_r1458081752)

packages/radix-ui-themes/src/components/data-list.tsx
kendallstrautman: [Another related comment](https://github.com/radix-ui/themes/pull/237/files#r1458138078)

packages/radix-ui-themes/src/components/data-list.props.ts
```diff
@@ -0,0 +1,105 @@
...
+    default: undefined,
+    responsive: true,
+  },
+  gapY: {
+    type: 'enum | string',
+    className: 'rt-r-rg',
+    customProperties: ['--row-gap'],
+    values: gapValues,
+    default: undefined,
+    responsive: true,
```
kendallstrautman, line 33: Decided to not abstract this out yet for flexibility in case we want to tweak the defaults easily.

packages/radix-ui-themes/src/components/data-list.css
```diff
@@ -0,0 +1,210 @@
...
+  position: relative;
+  box-sizing: border-box;
+  grid-column: span 2;
+}
+
+.rt-DataListLabel {
+  color: var(--gray-a11);
+  box-sizing: border-box;
+  display: block;
+  white-space: nowrap;
```
vladmoroz, line 16: Don't think we need `white-space: nowrap;`?

packages/radix-ui-themes/src/components/data-list.css
```diff
@@ -0,0 +1,210 @@
+.rt-DataListRoot {
+  display: grid;
+  grid-template-columns: auto 1fr;
+}
```
vladmoroz, line 4: Nit: can we include this only when the data list is horizontal?

The layout debugging experience is a bit weird otherwise

image

packages/radix-ui-themes/src/components/data-list.css
benoitgrelard: I personally think tokens should remain consistent, ie. when I use a space value of 3 it should always be the same resulting px value.

packages/radix-ui-themes/src/components/data-list.css
```diff
@@ -0,0 +1,210 @@
...
+      padding: 0;
+      gap: inherit;
+    }
+    &:where(.rt-r-dl-orient-vertical) :where(.rt-DataListLabel) {
+      white-space: normal;
+      padding: 0;
+      width: auto;
+    }
+  }
+}
```
vladmoroz, line 74: - Let's add all orientation-specific CSS in this block above `@breakpoints`, like that `grid-template-columns: auto 1fr`, so that all related CSS is compartmentalised
- Nit: similarly, I think it will be easier to read if this block is grouped by orientation rather than by part
- Nit: we'd usually do something generic like `.rt-r-orientation-horizontal` instead of classes scoped to components (`.rt-r-dl-...`)
- Could you add a fallback for when subgrid is not supported? (I assume `grid-template-columns: inherit` is what we want)
- Why do we need to set padding to 0?
- Why is `width: auto` needed?

packages/radix-ui-themes/src/components/data-list.props.ts
vladmoroz: I'd go ahead with abstracting it, safe to say we'll want all gap properties to work the same on all components

packages/radix-ui-themes/src/components/data-list.css
```diff
@@ -0,0 +1,210 @@
...
+.rt-DataListLabel {
+  color: var(--gray-a11);
+  box-sizing: border-box;
+  display: block;
+  white-space: nowrap;
+}
+
+.rt-DataListData {
+  box-sizing: border-box;
+  margin: 0;
```
vladmoroz, line 21: Let's put this margin reset into a related rule in `reset.css` and use the `rt-reset` class instead.
(e.g. check `link.tsx` as an example, `rt-reset` is applied separately to keep the component styles focused just on adding new styles)

packages/radix-ui-themes/src/components/data-list.css
```diff
@@ -0,0 +1,210 @@
+.rt-DataListRoot {
+  display: grid;
+  grid-template-columns: auto 1fr;
+}
+
+.rt-DataListItem {
+  position: relative;
+  box-sizing: border-box;
+  grid-column: span 2;
```
vladmoroz, line 9: - Same comment about including the CSS property conditionally based on the prop value
- Don't think we need `box-sizing: border-box` as there's no padding, also not sure about `position: relative`

packages/radix-ui-themes/src/components/data-list.css
```diff
@@ -0,0 +1,210 @@
...
+}
+
+.rt-DataListLabel {
+  color: var(--gray-a11);
+  box-sizing: border-box;
+  display: block;
```
vladmoroz, line 15: - Sounds like we need color and highContrast props instead of hardcoding this
- Is `display: block` and `box-sizing` needed?

packages/radix-ui-themes/src/components/data-list.props.ts
vladmoroz: I'd say we support all text sizes because if you put a CSS class on it manually it's going to work anyway

Size 9 is a bit stupid, but so are most size 9 things 🙂

packages/radix-ui-themes/src/components/data-list.css
vladmoroz: Let's reuse the existing width props as is because when it comes to using the space scale, "1", "2", etc. should always mean the same value, otherwise we'd invite trust issues around the prop values.

I do recognise this as an argument to review whether we need to expose space scale on all `width` props in other components at all. It's fair to say that in 90% of the cases the width values are too small to be useful on any component. Taking a note to sort it out.

vladmoroz (commented):
This is great, I hadn't realized how much better grid implementation would make it. Glad we don't need the pseudo-element alignment hack anymore too.

Leaving comments mostly just to clean up things a bit so they are inline with how other Radix Themes components are implemented.

packages/radix-ui-themes/src/components/data-list.props.ts
vladmoroz: @benoitgrelard I also wondered about this one, do you have any preference about how to name classes that have overlap with other utilities like this one?

packages/radix-ui-themes/src/components/data-list.tsx
```diff
@@ -0,0 +1,81 @@
...
+        <dl
+          {...dataListProps}
+          ref={forwardedRef}
+          className={classNames(className, 'rt-DataListRoot')}
+        />
+      </Text>
+    );
+  }
+);
+DataListRoot.displayName = 'DataListRootGrid';
```
vladmoroz, line 28: Any reason we'd use a different display name?

packages/radix-ui-themes/src/components/data-list.css
kendallstrautman: Yep! Good call.

packages/radix-ui-themes/src/components/data-list.tsx
kendallstrautman: Good catch, this was an oversight!

packages/radix-ui-themes/src/components/data-list.css
kendallstrautman: @vladmoroz I'm a bit confused because the largest width is ~64px that means majority of the width sizes wouldn't work for this use case. I get the need for consistency thought and perhaps most folks will just use the custom properties.

packages/radix-ui-themes/src/components/data-list.css
kendallstrautman: This was on the original component and since we don't set any default widths the labels get scrunched without it, but I'm fine to remove and folks can add it. 

Screenshot 2024-01-22 at 9 04 29 AM

packages/radix-ui-themes/src/components/data-list.css
kendallstrautman: @vladmoroz would you see `color` being set per label or at the Item or Root level?

packages/radix-ui-themes/src/components/data-list.css
vladmoroz: Yeah fair. I think that here we'll end up doing custom properties only—just want to look at it altogether everywhere width is used because it seems to be the general case. Let's just ignore the issue in this PR and follow up on the width prop after

packages/radix-ui-themes/src/components/data-list.css
vladmoroz: Definitely per label, you don't want the label and data to be always in sync

E.g. you might have to display an invalid or missing piece of data, and indicate that using a red label

packages/radix-ui-themes/src/components/data-list.tsx
```diff
@@ -0,0 +1,81 @@
...
+    );
+  }
+);
+DataListLabel.displayName = 'DataListLabel';
+
+type DataListDataElement = React.ElementRef<'dd'>;
+interface DataListDataProps extends React.ComponentPropsWithoutRef<'dd'> {}
+const DataListData = React.forwardRef<DataListDataElement, DataListDataProps>(
+  ({ children, className, ...props }, forwardedRef) => (
+    <dd {...props} ref={forwardedRef} className={classNames(className, 'rt-reset')}>
```
vladmoroz, line 64: Nit: even if not use in CSS, we'd usually keep a classname just so it's clear what is what when inspecting the HTML

```suggestion
    <dd {...props} ref={forwardedRef} className={classNames(className, 'rt-reset rt-DataListData')}>
```

vladmoroz (commented):
Looks awesome, we are almost there. Let's wrap it up with these comments and the color prop on the label.

packages/radix-ui-themes/src/components/data-list.props.ts
```diff
@@ -0,0 +1,53 @@
...
+    values: sizes,
+    default: '2',
+    responsive: true,
+  },
+  trim: {
+    ...trimProp,
+    className: 'rt-r-trim', // Custom trim styles due to grid layout
+  },
+} satisfies {
+  gap?: Omit<typeof gapProp.gap, 'default'> & { default: string };
```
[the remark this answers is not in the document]
benoitgrelard: We don't need to export, we can do `PropDef<(typeof gapProp.values)[number]>`.

packages/radix-ui-themes/src/components/data-list.css
kendallstrautman: For sure, I more was asking if we want to accept `color` higher and only apply it to the `Label`. But keeping it connected with the affected component makes sense to me.

packages/radix-ui-themes/src/components/data-list.props.ts
```diff
@@ -0,0 +1,53 @@
...
+    type: 'enum',
+    className: 'rt-r-size',
+    values: sizes,
+    default: '2',
+    responsive: true,
+  },
+  trim: {
+    ...trimProp,
+    className: 'rt-r-trim', // Custom trim styles due to grid layout
+  },
```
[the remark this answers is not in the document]
kendallstrautman: Good call, in my mind I was thinking alphabetical.

packages/radix-ui-themes/src/components/data-list.tsx
kendallstrautman: Word! I wasn't sure about that. Do you leave an empty class in the CSS file as well to connect that it exists?

packages/radix-ui-themes/src/components/flex.props.ts
```diff
@@ -1,13 +1,14 @@
...
 
 const displayValues = ['none', 'inline-flex', 'flex'] as const;
 const directionValues = ['row', 'column', 'row-reverse', 'column-reverse'] as const;
 const alignValues = ['start', 'center', 'end', 'baseline', 'stretch'] as const;
 const justifyValues = ['start', 'center', 'end', 'between'] as const;
 const wrapValues = ['nowrap', 'wrap', 'wrap-reverse'] as const;
-const gapValues = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'] as const;
 
 const flexPropDefs = {
+  ...gapProp,
```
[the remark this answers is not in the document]
kendallstrautman: Yes! My bad I saw some other components where we spread everything at the top.

packages/radix-ui-themes/src/components/data-list.tsx
vladmoroz: > Word! I wasn't sure about that. Do you leave an empty class in the CSS file as well to connect that it exists?

Don't think that we are consistent here, so no preference on my end

vladmoroz (approved):

kendallstrautman:
@vladmoroz [Updated examples here](https://themes-playground-git-ksadds-data-list-grid-subgrid-modulz.vercel.app/sink#datalist), lmk if that's more aligned!

vladmoroz:
Looks great @kendallstrautman, I'll go ahead and merge to rebase my PRs earlier

PR #238, opened 2024-01-17
Fix trim classname

kendallstrautman:
## Description

I believe this was a small bug from #235. See the classes [used in this file](https://github.com/radix-ui/themes/blob/main/packages/radix-ui-themes/src/styles/utilities/leading-trim.css) `rt-r-lt-*`


benoitgrelard (approved):
Yep! Well spotted!

PR #242, opened 2024-01-18
Add relative position to portalled content

kendallstrautman:
## Description

Add `position: relative` to components with portal content so that absolutely positioned children work naturally. 

## Testing steps

Reference the following components in [/sink](https://themes-playground-git-ksportalled-content-position-modulz.vercel.app/sink).
- AlertDialog
- Popover
- Dialog
- HoverCard
- Tooltip

## Notes

I didn't add this to _menu-type_ (Dropdown, Context, Select) components that use portal since the usecase doesn't seem as likely.

### DIFF
--- a/packages/radix-ui-themes/src/components/alert-dialog.css
+++ b/packages/radix-ui-themes/src/components/alert-dialog.css
@@ -2,4 +2,5 @@
 }
 
 .rt-AlertDialogContent {
+  position: relative;
 }
--- a/packages/radix-ui-themes/src/components/tooltip.css
+++ b/packages/radix-ui-themes/src/components/tooltip.css
@@ -2,6 +2,7 @@
   padding: var(--space-1) var(--space-2);
   background-color: var(--gray-12);
   border-radius: var(--radius-2);
+  position: relative;
 
   transform-origin: var(--radix-tooltip-content-transform-origin);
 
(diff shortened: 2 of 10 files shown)


packages/radix-ui-themes/src/components/tooltip.css
```diff
@@ -2,6 +2,7 @@
   padding: var(--space-1) var(--space-2);
   background-color: var(--gray-12);
   border-radius: var(--radius-2);
+  position: relative;
```
kendallstrautman, line 5: I was on the fence about the usecase for this one, assuming most folks are using text in tooltips. I could see potentially using an absolutely positioned icon though. Open to thoughts.

benoitgrelard: Yeah I can see what you mean, doesn't hurt to have it though.

benoitgrelard (approved):

packages/radix-ui-themes/src/components/alert-dialog.css
```diff
@@ -2,4 +2,5 @@
 }
 
 .rt-AlertDialogContent {
+  position: relative;
 }
```
vladmoroz, line 6: We don't need this style as Alert Dialog re-uses `rt-DialogContent` class

Should we delete the `alert-dialog.css` file altogether? Safe to say the design won't diverge from the base dialog

kendallstrautman: I can follow up with a cleanup PR 👍🏻

PR #243, opened 2024-01-18
Add support for `as` prop

kendallstrautman:
## Description

Add support for `as` prop to allow `div` or `span`. Previously these components could only be a `div` or use `asChild`.


packages/radix-ui-themes/src/components/box.tsx
```diff
@@ -14,16 +14,17 @@ interface BoxProps
     LayoutProps,
     BoxOwnProps {
   asChild?: boolean;
+  as?: 'span' | 'div';
```
kendallstrautman, line 17: 🤔 Should these be in the propDefs for docs generation?

packages/radix-ui-themes/changelog.md
```diff
@@ -29,10 +29,14 @@
   - Change the incorrect `display="block"` value to `display="initial"`
 - `Checkbox`, `RadioGroup`, `Switch`
   - Rework the internal HTML structure and styles. These components now render fewer HTML nodes and forward all props to the topmost node. Make sure that your code works as expected if you were relying on any of the implementation quirks to override styles or behaviour.
+- `Box`, `Flex`, `Grid`
+  - Add support for `as` prop to render as `span` or `div`
+- `Box`
```
benoitgrelard, line 34: can remove?

packages/radix-ui-themes/changelog.md
```diff
@@ -29,10 +29,14 @@
...
 - `Button`, `IconButton`
   - Add new `loading` prop
 - `Flex`
   - Add `gapX` and `gapY` props
+- `Grid`
```
benoitgrelard, line 39: Can remove?

benoitgrelard (commented):
For those components where we want to support both `asChild` but also `as` as a convenience, we need to alter the types a little bit so they are mutually exclusive and so that the attributes and event handlers are typed correctly.

A good example of that is the `Text` component.

Have a look at this as a starting point.

kendallstrautman:
🚧 Gotta fix the `ThemePanel` since it uses `typeof Box` in the interface which was previously static.

packages/radix-ui-themes/src/theme-panel.tsx
```diff
@@ -39,16 +40,16 @@ const ThemePanel = React.forwardRef<ThemePanelImplElement, ThemePanelProps>(
...
 
-type ThemePanelImplElement = React.ElementRef<typeof Box>;
-interface ThemePanelImplProps
-  extends React.ComponentPropsWithoutRef<typeof Box>,
-    ThemePanelImplPrivateProps {
-  onAppearanceChange?: (value: Exclude<ThemeOptions['appearance'], 'inherit'>) => void;
-}
+type ThemePanelImplElement = React.ElementRef<'div'>;
+// Narrow to only reference the `asChild` Box prop
+type ThemePanelImplOwnProps = CommonBoxProps & BoxAsChildProps;
```
kendallstrautman, line 45: I don't have the context as to why this extends the Box props, lmk if there's a better way to narrow here. I assume we don't want to support the `as` prop with this? If we do we'll need to add it to `Theme` as well.

packages/radix-ui-themes/src/components/box.tsx
```diff
@@ -2,31 +2,29 @@ import * as React from 'react';
...
-  extends React.ComponentPropsWithoutRef<'div'>,
-    MarginProps,
-    LayoutProps,
-    BoxOwnProps {
-  asChild?: boolean;
-}
+interface CommonBoxProps extends MarginProps, LayoutProps, BoxOwnProps {}
+type BoxAsChildProps = { asChild?: boolean; as?: never } & PropsWithoutRefOrColor<'div'>;
+type BoxSpanProps = { as?: 'span'; asChild?: never } & PropsWithoutRefOrColor<'span'>;
+type BoxDivProps = { as?: 'div'; asChild?: never } & PropsWithoutRefOrColor<'div'>;
```
benoitgrelard, line 14:
```suggestion
type BoxDivProps = { as?: 'div'; asChild?: never } & PropsWithoutRefOrColor<'div'>;
type BoxSpanProps = { as: 'span'; asChild?: never } & PropsWithoutRefOrColor<'span'>;
```

packages/radix-ui-themes/src/components/flex.tsx
```diff
@@ -4,25 +4,24 @@ import { Slot } from './slot';
...
-  extends React.ComponentPropsWithoutRef<'div'>,
-    MarginProps,
-    LayoutProps,
-    FlexOwnProps {
-  asChild?: boolean;
-}
+interface CommonFlexProps extends MarginProps, LayoutProps, FlexOwnProps {}
+type FlexAsChildProps = { asChild?: boolean; as?: never } & PropsWithoutRefOrColor<'div'>;
+type FlexSpanProps = { as?: 'span'; asChild?: never } & PropsWithoutRefOrColor<'span'>;
+type FlexDivProps = { as?: 'div'; asChild?: never } & PropsWithoutRefOrColor<'div'>;
```
benoitgrelard, line 14:
```suggestion
type FlexDivProps = { as?: 'div'; asChild?: never } & PropsWithoutRefOrColor<'div'>;
type FlexSpanProps = { as: 'span'; asChild?: never } & PropsWithoutRefOrColor<'span'>;
```

packages/radix-ui-themes/src/components/grid.tsx
```diff
@@ -4,25 +4,24 @@ import { Slot } from './slot';
...
-  extends React.ComponentPropsWithoutRef<'div'>,
-    MarginProps,
-    LayoutProps,
-    GridOwnProps {
-  asChild?: boolean;
-}
+interface CommonGridProps extends MarginProps, LayoutProps, GridOwnProps {}
+type GridAsChildProps = { asChild?: boolean; as?: never } & PropsWithoutRefOrColor<'div'>;
+type GridSpanProps = { as?: 'span'; asChild?: never } & PropsWithoutRefOrColor<'span'>;
+type GridDivProps = { as?: 'div'; asChild?: never } & PropsWithoutRefOrColor<'div'>;
```
benoitgrelard, line 14:
```suggestion
type GridDivProps = { as?: 'div'; asChild?: never } & PropsWithoutRefOrColor<'div'>;
type GridSpanProps = { as: 'span'; asChild?: never } & PropsWithoutRefOrColor<'span'>;
```

packages/radix-ui-themes/src/theme-panel.tsx
benoitgrelard: No we don't really as it stands, I'd be happy to remove that. I think we largely consider `ThemePanel` as just drop in with its own style that positions it, slides it in and out of the way etc.

packages/radix-ui-themes/src/components/box.tsx
vladmoroz, line 1: Do we want to add `display: block` to Box CSS so that `as="span"` doesn’t affect styling?

vladmoroz: or maybe not, I'm baffled here 🤔

kendallstrautman: 🤔 I assume that if folks want `as="span"` perhaps they don't want `display: block`? IMO it would be safe to assume they should handle that themselves 🤷🏻

kendallstrautman:
> I think we're fine with extending everything though

Sounds good, I'll try to just extend with `Flex`! I did get a type error before which is what pointed me towards this.

packages/radix-ui-themes/src/components/box.tsx
vladmoroz: We have a `display="inline"` option for setting the display value explicitly.

`as="span"` was rather requested a couple times as an easy way to set the tag for purist reasons, like when using the component inside `button` or `p`, where divs aren't supposed to be used.

My argument to set `display: block` in `box.css` is for consistency of behaviour with Flex and Grid, which are going to work the same way regardless of the tag. Curious what you and @benoitgrelard think.

packages/radix-ui-themes/src/components/box.tsx
kendallstrautman: Should the `as: 'div'` not be optional as well?

packages/radix-ui-themes/src/theme-panel.tsx
kendallstrautman: Cool that makes sense to me, I can remove it!

packages/radix-ui-themes/src/components/box.tsx
kendallstrautman: Oh I see what you mean now... yeah I'd say consistency makes sense to me if that's how Flex and Grid behave! Let me know if you want that on this PR or a separate one depending on Benoit's thoughts.

packages/radix-ui-themes/src/theme-panel.tsx
```diff
@@ -199,7 +197,6 @@ const ThemePanelImpl = React.forwardRef<ThemePanelImplElement, ThemePanelImplPro
             transitionTimingFunction: open ? 'ease-out' : 'ease-in',
             transform: open ? 'none' : 'translateX(105%)',
             boxShadow: open ? 'var(--shadow-5)' : 'var(--shadow-2)',
-            ...props.style,
```
kendallstrautman, line 202: Something to consider here, do we want to support the `style` prop? This was being provided by `Box` I assume.

vladmoroz: I personally wouldn't remove it and would pass through whatever `div` offers us but it doesn't matter really

kendallstrautman: I removed it since it was showing a type error as we removed the extension of `Box` props but I can adjust the types to pass it through.

kendallstrautman: Whoops. I realized I needed to add the `div` component props after removing `Box`. I'll add this back.

packages/radix-ui-themes/src/components/box.tsx
kendallstrautman: I pushed up that change 👌🏻

vladmoroz:
@kendallstrautman I fixed the conflicts caused by my recent PR and pushed into your branch

vladmoroz (approved):

benoitgrelard (approved):

PR #245, opened 2024-01-19
Add color prop to CheckboxItem and RadioItem

chsir17:
## Description

I've added the option to specify a color to `DropdownMenu.RadioItem`, `DropdownMenu.CheckboxItem`, `ContextMenu.RadioItem` and `ContextMenu.CheckboxItem`.

When I was using the library I found it unintuitive that you could specify a color the `DropdownMenu.Item` but not the the others. When reading the docs I thought I would be able to specify a color to these items also

## Testing steps

I've added examples in the sink: 

image

image

## Relates issues / PRs

None, decided to fix it myself instead of creating an issue

### DIFF
--- a/packages/radix-ui-themes/src/components/context-menu.tsx
+++ b/packages/radix-ui-themes/src/components/context-menu.tsx
@@ -9,6 +9,7 @@ import {
   contextMenuContentPropDefs,
   contextMenuItemPropDefs,
   contextMenuCheckboxItemPropDefs,
+  contextMenuRadioItemPropDefs,
 } from './context-menu.props';
 import { extractProps } from '../helpers';
 import { Theme, useThemeContext } from '../theme';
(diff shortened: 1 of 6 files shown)


vladmoroz (commented):
Looks good! Could you revert the sink page changes to keep the examples design realistic, and we'll merge that then?

packages/radix-ui-themes/src/components/context-menu.tsx
```diff
@@ -197,17 +206,24 @@ ContextMenuRadioItem.displayName = 'ContextMenuRadioItem';
...
 const ContextMenuCheckboxItem = React.forwardRef<
   ContextMenuCheckboxItemElement,
   ContextMenuCheckboxItemProps
 >((props, forwardedRef) => {
-  const { children, className, shortcut, ...itemProps } = props;
+  const {
+    children,
+    className,
+    shortcut,
+    color = contextMenuCheckboxItemPropDefs.color.default,
```
benoitgrelard, line 219: We should use the new extract function for those here right @vladmoroz?

vladmoroz: I remember I hadn't used it on the item because effectively it wasn't doing anything useful for us there anyway.

Let's settle this separately—maybe there's an argument for using it on all props everywhere just for consistency, but there are other components/parts like that.

chsir17:
> Looks good! Could you revert the sink page changes to keep the examples design realistic, and we'll merge that then?

@vladmoroz  Perfect! done: 66e91361b4fadaf865588dfd38988b44ae4d4c99

vladmoroz (approved):

PR #248, opened 2024-01-22
Add TabNav

benoitgrelard:
## Description

Adds a new `TabNav` components which looks visually the same as `Tabs` but is different semantically and usage-wise.

CleanShot 2024-01-22 at 12 05 46@2x

Fixes UI-2


packages/radix-ui-themes/src/components/base-tab-list.css
```diff
@@ -0,0 +1,140 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+.rt-BaseTabListLinkInner {
+  position: absolute;
+
+  :where(.rt-BaseTabListLink[data-state='inactive'], :not(.rt-TabNavLink[data-active])) & {
+    letter-spacing: var(--tabs-trigger-inactive-letter-spacing);
```
benoitgrelard, line 33: Not sure what to do about all the `--tabs-trigger-*` variables?

packages/radix-ui-themes/src/components/tabs.tsx
```diff
@@ -51,10 +51,10 @@ const TabsTrigger = React.forwardRef<TabsTriggerElement, TabsTriggerProps>(
       <TabsPrimitive.Trigger
         {...triggerProps}
         ref={forwardedRef}
-        className={classNames('rt-reset', 'rt-TabsTrigger', className)}
+        className={classNames('rt-reset', 'rt-BaseTabListLink', 'rt-TabsTRigger', className)}
```
vladmoroz, line 54:
```suggestion
        className={classNames('rt-reset', 'rt-BaseTabListLink', 'rt-TabsTrigger', className)}
```

packages/radix-ui-themes/src/components/base-tab-list.css
```diff
@@ -0,0 +1,140 @@
...
+  overflow-x: auto;
+  white-space: nowrap;
+
+  scrollbar-width: none;
+  &::-webkit-scrollbar {
+    display: none;
+  }
+}
+
+.rt-BaseTabListLink {
```
vladmoroz, line 12: Super nit: "link" in the class feels a bit specific, should the base call this an item or a trigger instead?

packages/radix-ui-themes/src/components/base-tab-list.css
```diff
@@ -0,0 +1,140 @@
...
+.rt-BaseTabListLinkInnerHidden {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+.rt-BaseTabListLinkInner {
+  position: absolute;
+
+  :where(.rt-BaseTabListLink[data-state='inactive'], :not(.rt-TabNavLink[data-active])) & {
```
vladmoroz, line 32: - Kinda bummer that we need to deal both with `data-state="active"` and `data-active`
- I believe that instead of `:not(.rt-TabNavLink[data-active])` we want `.rt-TabNavLink:not([data-active])`, otherwise you are selecting every element that is not an active tab link (vs every tab link that is not active)

packages/radix-ui-themes/src/components/tabs.tsx
```diff
@@ -36,7 +36,7 @@ const TabsList = React.forwardRef<TabsListElement, TabsListProps>((props, forwar
       data-accent-color={color}
       {...listProps}
       ref={forwardedRef}
-      className={classNames('rt-TabsList', className)}
+      className={classNames('rt-BaseTabList', 'rt-TabsList', className)}
```
vladmoroz, line 39: - Is it Tab List or Tabs List?
- Similarly, crossed my mind whether the components are:
  - Tabs and Tab Nav
  - Tabs and Tabs Nav

vladmoroz: OK I can see that "tablist" is a thing, and `Tabs.List` is probably incidental with how the original component is named Tabs

vladmoroz (commented):
Let's also update the changelog

packages/radix-ui-themes/src/components/base-tab-list.css
vladmoroz: Let's keep them as is

packages/radix-ui-themes/src/components/tab-nav.tsx
```diff
@@ -0,0 +1,90 @@
...
+  );
+}
+
+const TabNav = Object.assign(
+  {},
+  {
+    Root: TabNavRoot,
+    Item: TabNavItem,
+  }
+);
```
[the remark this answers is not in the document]
vladmoroz: Vibe check, Item vs Link

```jsx
<TabNav.Root>
  <TabNav.Item />
  <TabNav.Item />
  <TabNav.Item />
</TabNav.Root>
```

```jsx
<TabNav.Root>
  <TabNav.Link />
  <TabNav.Link />
  <TabNav.Link />
</TabNav.Root>
```

Looking at it like that, it feels like with the Link part, it is obvious that you need to supply the `href`. What do you think?

Also, curious what the `onSelect` callback on the original NavigationMenuLink is for? (https://www.radix-ui.com/primitives/docs/components/navigation-menu#link)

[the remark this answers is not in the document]
benoitgrelard: Happy to go with `Link` to make it more obvious.
`onSelect` is a handler where you can preventDefault to prevent the content to close.
Kinda irrelevant here, so we should omit from the TS definition probably.

packages/radix-ui-themes/src/components/tab-nav.tsx
```diff
@@ -0,0 +1,90 @@
...
+    MarginProps,
+    TabNavOwnProps {}
+const TabNavRoot = React.forwardRef<TabNavRootElement, TabNavRootProps>((props, forwardedRef) => {
+  const { children, className, color, ...rootProps } = extractProps(
+    props,
+    tabNavPropDefs,
+    marginPropDefs
+  );
+  return (
+    <NavigationMenu.Root
```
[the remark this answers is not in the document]
vladmoroz: Yeah absolutely, I had no expectation that this would expose anything else

packages/radix-ui-themes/src/components/tab-nav.tsx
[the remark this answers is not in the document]
vladmoroz: > Kinda irrelevant here, so we should omit from the TS definition probably.

Let's do this 👍

packages/radix-ui-themes/src/components/base-tab-list.css
vladmoroz: I think we are fine given that our alternatives are to either have two sets of variables for components with matching design, or to rename this set of variables into something that would work for both but then deal with this being a breaking change.

It's a bit unfortunate we haven't foreseen that we'd add another tabs component, but those tab nav links are also triggers in a way. I can take a note to return to this area again when we do a release with breaking changes. Unless you had something else in mind?

packages/radix-ui-themes/src/components/tab-nav.tsx
```diff
@@ -0,0 +1,90 @@
...
+  );
+});
+TabNavRoot.displayName = 'TabNavRoot';
+
+type TabNavItemElement = React.ElementRef<typeof NavigationMenu.Link>;
+interface TabNavItemProps extends React.ComponentPropsWithoutRef<typeof NavigationMenu.Link> {}
+const TabNavItem = React.forwardRef<TabNavItemElement, TabNavItemProps>((props, forwardedRef) => {
+  const { asChild = false, className, children, ...itemProps } = props;
+  return (
+    <NavigationMenu.Item className="rt-TabNavItem">
```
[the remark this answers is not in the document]
vladmoroz: Yeah seems OK, is there a scenario where you'd need access to the DOM nodes?

packages/radix-ui-themes/src/components/base-tab-list.css
```diff
@@ -0,0 +1,140 @@
...
+  align-items: center;
+  justify-content: center;
+}
+
+.rt-BaseTabListTriggerInner {
+  position: absolute;
+
+  :where(.rt-BaseTabListTrigger[data-state='inactive'], .rt-TabNavLink:not([data-active])) & {
+    letter-spacing: var(--tab-inactive-letter-spacing, var(--tabs-trigger-inactive-letter-spacing));
+    word-spacing: var(--tab-inactive-word-spacing, var(--tabs-trigger-inactive-word-spacing));
```
vladmoroz, line 34: I think we want to reverse the order of the variables here so that the new ones don't override any existing custom values

packages/radix-ui-themes/src/components/tab-nav.tsx
```diff
@@ -0,0 +1,92 @@
+'use client';
```
vladmoroz, line 1: Do we need 'use client'?

benoitgrelard: Yes we do, because `NavigationMenu` uses a bunch of client stuff (state, context, effects, etc).

packages/radix-ui-themes/src/components/base-tab-list.css
benoitgrelard: Ah yeah good point, switching…

vladmoroz (approved):

Issue #109, opened 2023-10-03
CSS specificity issues w/ tailwind


kevinmitch14:
I recently updated to v2 of Radix Themes and noticed that some of my styles were now being overridden by Tailwind.

**Notes**: This is only in prod deployments via Vercel, styles are applied correctly when locally developing

My setup was as follows:
```js
// layout.tsx
import "@/app/globals.css"; // Generic tailwind globals.css setup
import "@radix-ui/themes/styles.css";
...
```

When doing this, the tailwind base styles were taking priority over my radix themes styling. Both specificity (0,1,0).
<img width="500" alt="Screenshot 2023-10-03 at 13 27 06" src="https://github.com/radix-ui/themes/assets/55989505/baf67ea3-e9a2-40a9-a2f9-da8edd0938fb">

After playing around with some things, I have tried the following and now the styles are being applied correctly again.
```css
/* globals.css */
@tailwind base;
@tailwind components;
@tailwind utilities;
@import "@radix-ui/themes/styles.css";
...
```

To be honest I am not sure if this is the best option, but it has worked for me and may help others. It would be good to get some advice on this and best practices!

3den:
I am having the same problem, but moving the import to the the css file made no difference to me.

vladmoroz:
Those styles are a part of `@tailwind base`.

Import Radix Themes after `@tailwind base` and before other Tailwind styles.

```css
@tailwind base;
@import "@radix-ui/themes/styles.css";

@tailwind components;
@tailwind utilities;
```

kevinmitch14:
I'm aware of the problem so I opened this issue as a note for others. The behaviour is tricky because styles work as expected when developing locally. But when you push to prod, styles don't work.

It is not clear in the docs and I suspect that a lot of people will run into this same issue. 
> Import the global CSS file at the root of your application.
import '@radix-ui/themes/styles.css';

This code snippet does not suggest anything about importing in the CSS file. 

Happy to create a PR for the docs site

3den:
Changing the order of imports on the css file did not work for me, It worked only if i changed the order of imports in the layout.tsx (moving '@radix-ui/themes/styles.css' after all other css imports), but even that was not consistent and broke on prod.

vladmoroz:
> It is not clear in the docs and I suspect that a lot of people will run into this same issue.

Good call, we'll work out a guide on using Radix Themes with Tailwind.

[...]

PR #254, opened 2024-01-23
Update README.md (Tailwind Specificity Conflicts)

kevinmitch14:
Add note about tailwind specificity conflicts as there seems to be a lot of issues created in relation to this. Hopefully it will stop unnecessary issues until the fix is released!

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,10 @@
 
 **An open-source component library optimized for fast development, easy maintenance, and accessibility.**
 
+> [!NOTE]
+> **Usage with Tailwind**<br>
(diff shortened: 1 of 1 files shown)


kevinmitch14:
Maybe better to link to https://github.com/radix-ui/themes/issues/109#issuecomment-1747345743 as I saw @vladmoroz just mentioned this in response to an issue w/ tailwind.

vladmoroz:
Hey, thanks for looking into it. We decided to wait for a new Tailwind release first and add a note in the docs to upgrade it once it's out. For now I'll keep replying to the related issues and pointing to the comment with the resolution.

PR #256, opened 2024-01-24
Add RadioCardGroup

benoitgrelard:
## Description

https://github.com/radix-ui/themes/assets/1539897/58bc344f-5d5f-467e-985e-79605dfaa8e9

### DIFF
--- a/packages/radix-ui-themes/src/components/radio-card-group.css
+++ b/packages/radix-ui-themes/src/components/radio-card-group.css
@@ -70,7 +70,7 @@
     background-color: var(--color-surface);
     box-shadow: inset 0 0 0 1px var(--gray-a5);
 
-    &:where(:hover) {
+    &:where(:not([data-state='checked']):hover) {
       box-shadow: inset 0 0 0 1px var(--gray-a7);
     }
 
@@ -84,15 +84,6 @@
       outline-offset: 2px;
     }
   }
-
-  &:where(.rt-high-contrast) :where(.rt-RadioCardGroupItem) {
-    &:where([data-state='checked']) {
-      outline-color: var(--accent-12);
-    }
-    &:where(:focus-visible) {
-      outline-color: var(--accent-12);
-    }
-  }
 }
 
 /* classic */
@@ -104,14 +95,16 @@
     box-shadow: 0 0 0 1px var(--color-transparent), var(--shadow-2);
     transition: box-shadow 120ms;
 
-    &:where(:hover) {
-      transition-duration: 40ms;
-      box-shadow: var(--card-classic-hover-box-shadow);
-    }
+    &:where(:not([data-state='checked'])) {
+      &:where(:hover) {
+        transition-duration: 40ms;
+        box-shadow: var(--card-classic-hover-box-shadow);
+      }
 
-    &:where(:active) {
-      transition-duration: 40ms;
-      box-shadow: 0 0 0 1px var(--gray-a5), var(--shadow-2);
+      &:where(:active) {
+        transition-duration: 40ms;
+        box-shadow: 0 0 0 1px var(--gray-a5), var(--shadow-2);
+      }
     }
 
     &:where([data-state='checked']) {
@@ -128,11 +121,7 @@
 
 /* all high-contrast */
 
-.rt-RadioCardGroupRoot:where(.rt-high-contrast) :where(.rt-RadioCardGroupItem) {
-  &:where([data-state='checked']) {
-    outline-color: var(--accent-12);
-  }
-  &:where(:focus-visible) {
-    outline-color: var(--accent-12);
-  }
+.rt-RadioCardGroupRoot:where(.rt-high-contrast)
+  :where(.rt-RadioCardGroupItem[data-state='checked']) {
+  outline-color: var(--accent-12);
 }
--- a/packages/radix-ui-themes/src/components/radio-card-group.css
+++ b/packages/radix-ui-themes/src/components/radio-card-group.css
@@ -6,7 +6,6 @@
 .rt-RadioCardGroupItem {
   display: flex;
   align-items: center;
-  cursor: var(--cursor-radio);
 }
 
 /* Avoid unintentional drag interactions (e.g. on images) */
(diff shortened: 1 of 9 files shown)


packages/radix-ui-themes/src/components/radio-card-group.css
```diff
@@ -0,0 +1,132 @@
+.rt-RadioCardGroupRoot {
+  user-select: none;
+  cursor: default;
+}
+
+.rt-RadioCardGroupItem {
+  display: flex;
+  align-items: center;
+  cursor: var(--cursor-radio);
```
benoitgrelard, line 9: Do we want a separate cursor var for this component?

packages/radix-ui-themes/src/components/radio-card-group.css
```diff
@@ -0,0 +1,132 @@
...
+  pointer-events: none;
+}
+
+/***************************************************************************************************
+ *                                                                                                 *
+ * SIZES                                                                                           *
+ *                                                                                                 *
+ ***************************************************************************************************/
+
+.rt-RadioCardGroupItem {
```
benoitgrelard, line 23: Changes from the WorkDS version:
- I removed the fixed height and rely on padding instead
- padding and radius have been aligned with the card ones for the same 3 sizes

packages/radix-ui-themes/src/components/radio-card-group.css
```diff
@@ -0,0 +1,132 @@
...
+
+    &:where([data-state='checked']) {
+      outline: 2px solid var(--accent-9);
+      outline-offset: -1px;
+    }
+
+    &:where(:focus-visible) {
+      outline: 2px solid var(--accent-9);
+      outline-offset: -1px;
+      box-shadow: 0 0 0 4px var(--accent-a6);
```
benoitgrelard, line 74: I used a lower step here for more contrast (compared to the WorkDS one)
I think it works better.

I'm also generally questioning whether this style is fine or not, we don't really have it anywhere else in Radix Themes.

packages/radix-ui-themes/src/components/radio-card-group.css
```diff
@@ -0,0 +1,132 @@
...
+    }
+  }
+
+  &:not(:where(.rt-high-contrast)) :where(.rt-RadioCardGroupItem):where([data-state='checked']) {
+    background-color: var(--accent-a2);
+    color: var(--accent-12);
```
benoitgrelard, line 80: Only doing these for non-high-contrast, looked odd otherwise.

packages/radix-ui-themes/src/components/radio-card-group.css
```diff
@@ -0,0 +1,132 @@
...
+      outline-color: var(--accent-12);
+    }
+  }
+}
+
+/* classic */
+
+.rt-RadioCardGroupRoot:where(.rt-variant-classic) {
+  :where(.rt-RadioCardGroupItem) {
+    background-color: var(--color-panel);
```
benoitgrelard, line 96: Not sure what the background should be for classic, should it remain `--color-surface` and just differ by the shadow?

packages/radix-ui-themes/src/components/radio-card-group.css
```diff
@@ -0,0 +1,132 @@
...
+
+    &:where(:hover) {
+      transition-duration: 40ms;
+      box-shadow: var(--card-classic-hover-box-shadow);
+    }
+
+    &:where(:active) {
+      transition-duration: 40ms;
+      box-shadow: 0 0 0 1px var(--gray-a5), var(--shadow-2);
+    }
```
benoitgrelard, line 109: I'm reusing the shadow treatments from `Card`.

benoitgrelard (commented):
I still need to do some work on the explore components page to present that component but thought I'd put this up early for feedback.

vladmoroz:
Checking the design, let's do this for all variants:
- No background color change in the active state
- `-2px` active state outline offset so it is level with the card
- Use focus technique like on the switch/slider thumbs. Looks slightly uglier but consistent with what we have, which seems to be the right trade-off given it's just this one component.

packages/radix-ui-themes/src/components/radio-card-group.css
vladmoroz: Nope, I'd definitely a button and not radio cursor, that would have to be the same when you have a Card button

packages/radix-ui-themes/src/components/radio-card-group.css
```diff
@@ -0,0 +1,138 @@
...
+
+/* all high-contrast */
+
+.rt-RadioCardGroupRoot:where(.rt-high-contrast) :where(.rt-RadioCardGroupItem) {
+  &:where([data-state='checked']) {
+    outline-color: var(--accent-12);
+  }
+  &:where(:focus-visible) {
+    outline-color: var(--accent-12);
+  }
```
vladmoroz, line 137: This doesn't work, but we don't really need it—don't think we adjust the focus outline color on the high contrast variants anymore

packages/radix-ui-themes/src/components/radio-card-group.css
```diff
@@ -0,0 +1,138 @@
...
+
+  &:where(.rt-high-contrast) :where(.rt-RadioCardGroupItem) {
+    &:where([data-state='checked']) {
+      outline-color: var(--accent-12);
+    }
+    &:where(:focus-visible) {
+      outline-color: var(--accent-12);
+    }
+  }
```
vladmoroz, line 95: Repeated high-contrast?

vladmoroz:
Should we remove hover and pressed styles from the classic variant when it's checked?

packages/radix-ui-themes/src/components/radio-card-group.css
benoitgrelard: Yep, that's a mistake, wasn't trying to do that anyway as we don't adjust it like you say.

packages/radix-ui-themes/src/components/radio-card-group.css
```diff
@@ -0,0 +1,127 @@
...
+.rt-RadioCardGroupItem {
+  position: relative;
+  /* pseudo element for focus outline */
+  &::before {
+    content: '';
+    display: block;
+    position: absolute;
+    inset: 0;
+    border-radius: inherit;
+  }
```
vladmoroz, line 63: Super nit, but it drives me insane that when trying to target the radio cards with inspector, the pseudo catches pointer events

benoitgrelard: Yeah hate that too, nice catch!

vladmoroz:
@benoitgrelard I'm happy with the layout/sizes now, let me know if my changes make sense
https://github.com/radix-ui/themes/pull/256/commits/ed7fc0a6a0592c06ac113c4952f044d10ea51560

vladmoroz (approved):

Issue #181, opened 2023-12-11
[Feature request] Allow tabs to wrap to next line


penx:
It would be useful if we could optionally make the Tabs List component wrap containing Trigger items to the next line, rather than requiring a horizontal scroll.

vladmoroz:
You could set `white-space: normal` on `TabsList`. What does your design look like so we can consider whether it makes sense in the component API?

penx:
I don't have a design from our current app as we've been avoiding using Tabs for now, partly due to this.

Thanks for your suggestion re:white space.

As an example design, say you have the following:

<img width="1183" alt="Screenshot 2024-01-17 at 10 48 30" src="https://github.com/radix-ui/themes/assets/1027024/9b14ecf2-9715-426b-a4b3-2d8a3d97f97d">

In the mobile view, on the left, the user cannot see the "Settings" tab, and it is not clear that this is a scrollable area.

Wrapping the tabs to the next line would be preferred, from my perspective.

This follows, for example, what the tabs here do (Get Started/Styles/Components):

https://design-system.service.gov.uk/

Alternatively in the actual govuk tabs component, tabs turn in to a bulleted list when there is not enough horizontal space to fit them all, though I expect this is a manually set breakpoint rather than calculated at runtime:

https://design-system.service.gov.uk/components/tabs/

penx:
Ok we have implemented this in our designs as per your suggestion on overriding CSS on TabsList.

In the end we had to do this:


```ts
<TabsList style={{ rowGap: 'var(--space-2)', height: 'auto' }}>
```

<img width="370" alt="Screenshot 2024-01-24 at 11 24 51" src="https://github.com/radix-ui/themes/assets/1027024/27a5111e-0685-48c8-9ea9-a475ecd7a976">
<img width="447" alt="Screenshot 2024-01-24 at 11 24 40" src="https://github.com/radix-ui/themes/assets/1027024/0ac3d089-24e9-4e03-8309-28e754a215cd">

vladmoroz:
Ah I see, that is reasonable. We'll accommodate that.

PR #257, opened 2024-01-24
Add justify and wrap props to Tabs

vladmoroz:
Closes https://github.com/radix-ui/themes/issues/181

### DIFF
--- a/packages/radix-ui-themes/src/components/base-tab-list.props.ts
+++ b/packages/radix-ui-themes/src/components/base-tab-list.props.ts
@@ -1,13 +1,31 @@
 import { PropDef, colorProp, highContrastProp } from '../helpers';
 
 const sizes = ['1', '2'] as const;
+const wrapValues = ['nowrap', 'wrap', 'wrap-reverse'] as const;
+const justifyValues = ['start', 'center', 'end'] as const;
 
 const baseTabListPropDefs = {
   size: { type: 'enum', className: 'rt-r-size', values: sizes, default: '2', responsive: true },
+  wrap: {
+    type: 'enum',
+    className: 'rt-r-fw',
+    values: wrapValues,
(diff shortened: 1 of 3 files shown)


packages/radix-ui-themes/src/components/base-tab-list.props.ts
```diff
@@ -1,13 +1,31 @@
 import { PropDef, colorProp, highContrastProp } from '../helpers';
 
 const sizes = ['1', '2'] as const;
+const wrapValues = ['nowrap', 'wrap', 'wrap-reverse'] as const;
+const justifyValues = ['start', 'center', 'end'] as const;
```
kendallstrautman, line 5: Is 'between' not useful for this case? Wondering about consistency with `Flex`

kendallstrautman (approved):
LGTM, just one question on the justify values.

packages/radix-ui-themes/src/components/base-tab-list.props.ts
vladmoroz: Yeah, I decided to omit `between` because it looked unreasonably bad in the context of tabs

packages/radix-ui-themes/src/components/base-tab-list.props.ts
```diff
@@ -1,13 +1,31 @@
...
+    default: undefined,
+    responsive: true,
+  },
+  justify: {
+    type: 'enum',
+    className: 'rt-r-jc',
+    values: justifyValues,
+    default: 'start',
+    responsive: true,
+  },
```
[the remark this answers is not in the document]
vladmoroz: I'll extract `wrap` into a separate file, justify will probably stay as is as it doesn't have `between` and doesn't need that function to parse the value

PR #259, opened 2024-01-25
Improve how accent color and high contrast work when components are nested

vladmoroz:
- Improve how accent color and high contrast work when components are nested
  - For components like Text and Heading, `--accent-12` is going to be resolved as long as there is some parent component with `data-accent-color` in the DOM that is not `.radix-themes`
- Make it so that `<Code variant="ghost">` color works just like Text, inheriting by default, but colored when an explicit `color` prop is passed
- Simplify styles for Code and Link combinations

Before:
https://themes-playground-jnjh1ct8k-modulz.vercel.app/test-high-contrast

After:
https://themes-playground-git-vlad-ghost-code-color-modulz.vercel.app/test-high-contrast

### DIFF
--- a/packages/radix-ui-themes/src/components/code.css
+++ b/packages/radix-ui-themes/src/components/code.css
@@ -10,6 +10,10 @@
     var(--code-letter-spacing) + var(--letter-spacing, var(--default-letter-spacing))
   );
   border-radius: calc((0.5px + 0.2em) * var(--radius-factor));
+
+  & :where(&) {
+    font-size: inherit;
+  }
 }
 
 /***************************************************************************************************
@@ -78,24 +82,14 @@
 
 .rt-Code:where(.rt-variant-ghost) {
   --code-variant-font-size-adjust: var(--code-font-size-adjust);
-  color: var(--accent-a11);
 
-  &:where(.rt-high-contrast) {
+  &:where([data-accent-color]) {
+    color: var(--accent-a11);
+  }
+
+  &:where([data-accent-color].rt-high-contrast),
+  :where([data-accent-color]:not(.radix-themes)) &:where(.rt-high-contrast) {
     color: var(--accent-12);
-  }
-
-  :where(.rt-Link:not(.rt-underline-hover)) & {
-    text-decoration-line: underline;
-    text-decoration-color: inherit;
-    text-decoration-thickness: inherit;
-  }
-
-  :where(.rt-Link:has(&:only-child)) {
-    &:where(:focus-visible) .rt-Code {
-      text-decoration: none;
-      outline: 2px solid var(--color-focus-root);
-      outline-offset: 2px;
-    }
   }
 }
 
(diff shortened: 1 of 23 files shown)


packages/radix-ui-themes/src/components/code.css
```diff
@@ -116,17 +110,17 @@
...
-          background-color: var(--accent-10);
-        }
-        &:where(.rt-high-contrast:hover) {
-          background-color: var(--accent-12);
-          /* Re-use base button hover filter */
-          filter: var(--base-button-solid-high-contrast-hover-filter);
-        }
+  :where(.rt-Link) & {
+    /* Create a new stacking context (otherwise, `filter` may do it on hover) */
+    isolation: isolate;
```
vladmoroz, line 115: Fixes a tiny edge case I saw

https://github.com/radix-ui/themes/assets/8441036/30ef31db-cbf4-4c40-a296-e58662013dbe

packages/radix-ui-themes/src/helpers/props/color.prop.ts
vladmoroz, line 1: I think I'm directionally right about the outcome but not sure if there is a better way to implement. Essentially:
- Most components will use `colorProp` and get an empty `data-accent-color` attribute if you don't pass an explicit `color`. This defines the right high-contrast colors for Text, Heading, Link, etc. without having to whitelist which components have to do that. As the result, a couple of new component combinations would now work out of the box, e.g. like this:
  ```jsx
  <Badge color="gray">
    Status: <Text highContrast>Backlog</Text>
  </Badge>
  ```
- Components like Text, Heading, Link that need to be able to derive their color from a parent without setting own color would use `inheritedColorProp`

packages/radix-ui-themes/src/helpers/props/color.prop.ts
```diff
@@ -2,6 +2,12 @@ import { themePropDefs } from '../../theme-options';
 import type { PropDef } from '..';
 
 const colorProp = {
```
vladmoroz, line 4: I think I'm directionally right about the outcome but not sure if there is a better way to implement. Essentially:
- Most components will use `colorProp` and get an empty `data-accent-color` attribute if you don't pass an explicit `color`. This defines the right high-contrast colors for Text, Heading, Link, etc. without having to whitelist which components have to do that. As the result, a couple of new component combinations would now work out of the box, e.g. like this:
  ```jsx
  <Badge color="gray">
    Status: <Text highContrast>Backlog</Text>
  </Badge>
  ```
- Components like Text, Heading, Link that need to be able to derive their color from a parent when their own color is not set explicitly would use `inheritedColorProp`

packages/radix-ui-themes/src/components/code.css
```diff
@@ -10,6 +10,10 @@
     var(--code-letter-spacing) + var(--letter-spacing, var(--default-letter-spacing))
   );
   border-radius: calc((0.5px + 0.2em) * var(--radius-factor));
+
+  & :where(&) {
+    font-size: inherit;
+  }
```
vladmoroz, line 16: All the formatting inline tags are technically nestable within each other

benoitgrelard (approved):

benoitgrelard:
Just some conflict to fix

PR #260, opened 2024-01-25
Improve string union prop types

vladmoroz:


packages/radix-ui-themes/src/helpers/union.ts
```diff
@@ -0,0 +1,3 @@
+// Creates a union type of string literals with strings, but retains intellisense for the literals.
+// Union<string, 'foo' | 'bar'> => string | 'foo' | 'bar'
```
benoitgrelard, line 2:
```suggestion
// Union<string, 'foo' | 'bar'> => string | Omit<string, 'foo' | 'bar'>
```

packages/radix-ui-themes/src/helpers/responsive.ts
```diff
@@ -8,7 +8,7 @@ type Responsive<T> = T | Partial<Record<Breakpoints, T>>;
 interface GetResponsiveStylesOptions {
   className: string;
   customProperties: `--${string}`[];
-  value: Responsive<StringOrValue<string>> | Responsive<string> | undefined;
+  value: Responsive<Union<string, string>> | Responsive<string> | undefined;
```
benoitgrelard, line 11: These all read weird now, why are they even needed?

Surely `Union<string, string>` is the same as `string` no? So then that'd become `Responsive<string>` which is already in the union next…

vladmoroz: Technically both `StringOrValue<string>` and `Union<string, string>` is the same as `string`, but TS doesn’t seem to recognize that and complains

I added defaults to the Union type so at least it doesn't read weird, what do you think?
https://github.com/radix-ui/themes/pull/260/commits/42890e6a3dfb065fab8c75c05ed50c5e2d9981c9

packages/radix-ui-themes/src/helpers/props/margin.props.ts
```diff
@@ -5,7 +5,7 @@ import { mergeStyles } from '../merge-styles';
 import type { PropDef, GetPropDefTypes } from './prop-def';
 
 // prettier-ignore
-const marginValues = ['auto', '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '-1', '-2', '-3', '-4', '-5', '-6', '-7', '-8', '-9'] as const;
```
vladmoroz, line 8: Noticed I had forgotten to remove auto

benoitgrelard (approved):

PR #261, opened 2024-01-26
Confine dialog scroll to viewport

vladmoroz:
See the [test-dialog](https://themes-playground-git-vlad-dialog-scroll-modulz.vercel.app/test-dialog) page

Before

https://github.com/radix-ui/themes/assets/8441036/f142a81b-295a-4776-80b6-164a17b14a53

After

https://github.com/radix-ui/themes/assets/8441036/037bd2eb-caed-4743-ab97-d3207ae8573e

### DIFF
--- a/packages/radix-ui-themes/src/components/dialog.css
+++ b/packages/radix-ui-themes/src/components/dialog.css
@@ -1,46 +1,43 @@
 .rt-DialogOverlay {
-  display: flex;
-  align-items: center;
-  justify-content: center;
   position: fixed;
   inset: 0;
 
-  --dialog-overlay-padding-top: var(--space-4);
-  --dialog-overlay-padding-bottom: max(var(--space-4), 4vh);
-  padding-top: var(--dialog-overlay-padding-top);
-  padding-bottom: var(--dialog-overlay-padding-bottom);
-  padding-left: var(--space-4);
-  padding-right: var(--space-4);
-
-  &::after {
+  &::before {
+    position: fixed;
     content: '';
-    position: absolute;
     inset: 0;
     background-color: var(--color-overlay);
   }
 }
 
+.rt-DialogScroll {
+  display: flex;
+  overflow: auto;
+  position: absolute;
+  inset: 0;
+}
+
+.rt-DialogScrollPadding {
+  margin: auto;
+  padding-top: var(--space-6);
+  padding-bottom: max(var(--space-6), 6vh);
+  padding-left: var(--space-4);
+  padding-right: var(--space-4);
+}
+
 .rt-DialogContent {
   width: 100%;
   max-width: 580px;
-  outline: none;
-  overflow: auto;
-  background-color: var(--color-panel-solid);
-  box-shadow: var(--shadow-6);
-  box-sizing: border-box;
   z-index: 1;
   position: relative;
 
   --inset-padding: var(--dialog-content-padding);
   padding: var(--inset-padding);
+  box-sizing: border-box;
 
-  /* prettier-ignore */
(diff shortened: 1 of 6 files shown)


packages/radix-ui-themes/src/components/dialog.tsx
```diff
@@ -41,11 +41,15 @@ const DialogContent = React.forwardRef<DialogContentElement, DialogContentProps>
...
-            />
+            <div className="rt-DialogScroll">
+              <div className="rt-DialogScrollPadding">
+                <DialogPrimitive.Content
+                  {...contentProps}
+                  ref={forwardedRef}
+                  className={classNames('rt-DialogContent', className)}
+                />
+              </div>
+            </div>
```
vladmoroz, line 52: Instead of 2 DOM elements, we could achieve it with just 1 more DOM element, or potentially with no extra DOM elements if we distributed some styles onto the pseudo elements, however I found that it adds a layer of indirection in the styles and when inspecting the resulting DOM that I wasn't sure worth it.

packages/radix-ui-themes/src/components/dialog.css
```diff
@@ -1,46 +1,43 @@
...
-  --dialog-overlay-padding-top: var(--space-4);
-  --dialog-overlay-padding-bottom: max(var(--space-4), 4vh);
-  padding-top: var(--dialog-overlay-padding-top);
-  padding-bottom: var(--dialog-overlay-padding-bottom);
-  padding-left: var(--space-4);
-  padding-right: var(--space-4);
-
-  &::after {
+  &::before {
+    position: fixed;
```
vladmoroz, line 6: Tried to avoid having a pseudo and use background-color on `.rt-DialogOverlay` directly and animate that, but there's some bug in Chrome when the animation to open the dialog is butchered on the initial page load (and works fine when you open the dialog again)

packages/radix-ui-themes/src/components/dialog.css
```diff
@@ -1,46 +1,43 @@
...
+  inset: 0;
+}
+
+.rt-DialogScrollPadding {
+  margin: auto;
+  padding-top: var(--space-6);
+  padding-bottom: max(var(--space-6), 6vh);
+  padding-left: var(--space-4);
+  padding-right: var(--space-4);
+}
```
vladmoroz, line 26: An extra element is needed because padding-bottom doesn't add any bottom padding in overflowing scroll containers

kendallstrautman (commented):
Should we make the behavior prop driven? Container scroll vs. viewport scroll? Thinking of situations where I want to contain the dialog content to a certain height and I do want to scroll within the container. In my limited playing with this, if I set height on `DialogContent` the text overflows. 

Screenshot 2024-01-26 at 12 54 02 PM

vladmoroz:
@kendallstrautman can you think of a case where the UX would be a better fit with scroll confined to the dialog content?

Looking at your example, you might just need an overflow: auto there to re-create the previous behaviour

kendallstrautman:
@vladmoroz I'm not sure if the UX is better with a fixed height dialog, but it's a pattern I do see a lot where someone wants a dialog to feel fixed and not scroll on the rest of the page. I don't think its a blocker, just noting that if folks do have just `height` set, they'll need to adjust somethings to retain that behavior with the new release.

kendallstrautman (approved):
Works well for the intended usecase!

vladmoroz:
Interesting, yeah I’d assume that design is more common because it's just trivial to implement compared to the full-height scroll which comes with a few gotchas.

Anyway, doesn't cost us anything to also add `overflow: auto` on the dialog content to allow fixed-height dialogs, added that:

https://github.com/radix-ui/themes/assets/8441036/6c045599-5237-486a-93f9-edb25a6b82be

PR #263, opened 2024-01-29
Fix TextArea bugged by the Grammarly extension and add resize prop 

### LINKED ISSUE #167: TextArea style bug on Chrome
I have an issue where the TextArea outline is shown on the first render then when I click inside the outline disappears.

First render
image

focus
image

clicking outside 
image

### LINKED ISSUE #223: `TextArea` should be resizable
resizeplease

This libray has very nice visuals, but this should be fixed for better accessibilty. Note that adding `resize: 'vertical'` to the property of `TextArea` didn't work. If it's not available, I'd like to know why it can't be supported.

vladmoroz:
Closes #167
Closes #223


packages/radix-ui-themes/src/components/text-area.css
```diff
@@ -27,7 +48,8 @@
   padding: var(--text-area-padding-y) var(--text-area-padding-x);
 
   &:where(:focus) {
-    outline: 2px solid var(--color-focus-root);
+    outline: 1px solid;
+    outline-color: inherit;
```
vladmoroz, line 52: Ended up having to maintain a focus outline both on the wrapper and on the actual text area input element

- Since our text area element comes in a wrapper div for layout reasons, we have to put `resize` on the actual div so that resizing horizontally works as expected
- To make `resize` work on the div, it needs `overflow: hidden`
- `overflow: hidden` trims the outer part of the focus outline of the text area input element
- Because of that, the focus outline needs to go on the wrapper
- However, `.rt-TextAreaChrome` box shadow goes on top of the inner part of the focus outline, hence we need to do the inner focus outline separately 

To visualize these problems...

2px outline just on the wrapper:

Screenshot 2024-01-29 at 10 10 47

1px outer outline on the wrapper and 1px inner outline on the text area element

Screenshot 2024-01-29 at 10 10 17

2px outline on the wrapper and 1px inner outline on the text area element (this version)

Screenshot 2024-01-29 at 10 10 43

***

I think that we'll be able to shed a good part of the text area and text input CSS when the usage of Firefox versions that don't support `:has` fades away a bit more, for now this should be a good practical version.

benoitgrelard (commented):
Works well, my only issue with it I think is the somewhat non-standard icon.
I feel like it's not obvious what the circle is…

vladmoroz:
@benoitgrelard all right, removed resizer styles

benoitgrelard (approved):

PR #264, opened 2024-01-29
Add interactive Kbd styles

vladmoroz:
Something we had to hack together on the spot in the WorkOS docs here

image

### DIFF
--- a/packages/radix-ui-themes/src/components/card.css
+++ b/packages/radix-ui-themes/src/components/card.css
@@ -139,7 +139,18 @@
 
   &:where(button, [href]) {
     @media (hover: hover) {
-      &:where(:hover)::after {
+      &:where(:hover) {
+        &::after {
+          box-shadow: 0 0 0 1px var(--gray-a7);
+
+          @supports (box-shadow: 0 0 0 1px color-mix(in oklab, white, black)) {
+            box-shadow: 0 0 0 1px color-mix(in oklab, var(--gray-a7), var(--gray-7) 25%);
+          }
+        }
+      }
+    }
+    &:where([data-state='open']) {
+      &::after {
         box-shadow: 0 0 0 1px var(--gray-a7);
 
         @supports (box-shadow: 0 0 0 1px color-mix(in oklab, white, black)) {
@@ -223,6 +234,12 @@
         }
       }
     }
+    &:where([data-state='open']) {
+      &::after {
+        transition-duration: 40ms;
+        box-shadow: var(--card-classic-hover-box-shadow);
+      }
+    }
     &:where(:active:not([data-state='open'])) {
       &::after {
         transition-duration: 40ms;
@@ -245,6 +262,9 @@
       &:where(:hover) {
         background-color: var(--gray-a3);
       }
+    }
+    &:where([data-state='open']) {
+      background-color: var(--gray-a3);
     }
     &:where(:active:not([data-state='open'])) {
       background-color: var(--gray-a4);
--- a/packages/radix-ui-themes/src/components/code.css
+++ b/packages/radix-ui-themes/src/components/code.css
@@ -1,6 +1,3 @@
-/* stylelint-disable selector-max-type */
-/* Disable selector-max-type rule to target individual element types. */
-
 .rt-Code {
   --code-variant-font-size-adjust: calc(var(--code-font-size-adjust) * 0.95);
   box-sizing: border-box;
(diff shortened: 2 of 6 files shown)


packages/radix-ui-themes/src/components/code.css
```diff
@@ -1,3 +1,6 @@
+/* stylelint-disable selector-max-type */
+/* Disable selector-max-type rule to target individual element types. */
```
vladmoroz, line 2: Getting close to disabling this rule in the global config and relying on a verbal convention not to use type selector without `:where`

packages/radix-ui-themes/src/components/code.css
```diff
@@ -110,7 +113,8 @@
     color: var(--accent-12);
   }
 
-  :where(.rt-Link) & {
+  :where(.rt-Link) &,
+  &:where(a, button) {
```
vladmoroz, line 117: Since we already have styles for interactive Code, should we support `<Code asChild>` into buttons/links?

packages/radix-ui-themes/src/components/card.css
```diff
@@ -139,7 +139,18 @@
...
+          box-shadow: 0 0 0 1px var(--gray-a7);
+
+          @supports (box-shadow: 0 0 0 1px color-mix(in oklab, white, black)) {
+            box-shadow: 0 0 0 1px color-mix(in oklab, var(--gray-a7), var(--gray-7) 25%);
+          }
+        }
+      }
+    }
+    &:where([data-state='open']) {
+      &::after {
```
vladmoroz, line 153: Added missing [data-state='open'] styles as we usually pair them with all :hover styles

packages/radix-ui-themes/src/components/kbd.props.ts
```diff
@@ -10,8 +10,13 @@ const kbdPropDefs = {
     default: undefined,
     responsive: true,
   },
+  asChild: {
+    type: 'boolean',
+    default: undefined,
+  },
```
vladmoroz, line 16: I think we should start adding asChild in the prop defs on every component here so that it shows up in the docs. I'll do a more holistic pass on asChild for other components too.

benoitgrelard: To be clear, it does show in the docs, but we inject that prop def manually I believe.

packages/radix-ui-themes/src/components/kbd.tsx
```diff
@@ -4,14 +4,20 @@ import { kbdPropDefs } from './kbd.props';
 import { extractProps, GetPropDefTypes, marginPropDefs } from '../helpers';
 
 import type { MarginProps } from '../helpers';
+import { Slot } from '@radix-ui/react-slot';
```
benoitgrelard, line 7: nit:

```suggestion
import { Slot } from '@radix-ui/react-slot';
import { extractProps, GetPropDefTypes, marginPropDefs } from '../helpers';

import type { MarginProps } from '../helpers';
```

packages/radix-ui-themes/src/components/kbd.tsx
```diff
@@ -4,14 +4,20 @@ import { kbdPropDefs } from './kbd.props';
...
-const Kbd = React.forwardRef<KbdElement, KbdProps>((props, forwardedRef) => {
-  const { className, ...kbdProps } = extractProps(props, kbdPropDefs, marginPropDefs);
-  return <kbd {...kbdProps} ref={forwardedRef} className={classNames('rt-Kbd', className)} />;
-});
+const Kbd = React.forwardRef<KbdElement, KbdProps>(
+  ({ asChild = false, ...props }, forwardedRef) => {
+    const Comp = asChild ? Slot : 'kbd';
+    const { className, ...kbdProps } = extractProps(props, kbdPropDefs, marginPropDefs);
+    return (
+      <Comp {...kbdProps} ref={forwardedRef} className={classNames('rt-reset rt-Kbd', className)} />
```
benoitgrelard, line 17: nit:

```suggestion
      <Comp {...kbdProps} ref={forwardedRef} className={classNames('rt-reset', 'rt-Kbd', className)} />
```

benoitgrelard (approved):

packages/radix-ui-themes/src/components/kbd.props.ts
vladmoroz: Gotcha, I just checked and we do it like this

image

Thought we don't because there has been a bunch of bug reports about missing and/or not working `asChild`

Would you agree it should come from the prop defs?

benoitgrelard: I think it would make sense to ultimately come from prop defs.

PR #265, opened 2024-01-29
Split CSS exports

vladmoroz:
Split the CSS exports into:
- tokens.css
- components.css
- utilities.css

The existing `styles.css` file remains as is.

This is mainly so that consumers can import `utilities.css` on top of their custom styles. In WorkOS we have a problem when our custom components styles would overwrite the utilities, making it impossible to create new components that work with the provided utility classes.

No changelog yet, will want to see how we end up using this before we can document importing separate files as something we recommend.


packages/radix-ui-themes/src/styles/utilities/index.css
```diff
@@ -0,0 +1,37 @@
...
+@import './height.css';
+@import './min-height.css';
+@import './max-height.css';
+@import './inset.css';
+@import './justify-content.css';
+@import './leading-trim.css';
+@import './margin.css';
+@import './padding.css';
+@import './position.css';
+/* @import './resize.css'; */
```
vladmoroz, line 29: When text area PR merges...

packages/radix-ui-themes/src/components/popper.css
vladmoroz, line 1: Treating this as component styles essentially, any better ideas?

benoitgrelard: Yes that makes sense as it's shared CSS code used by components.
Not sure about the rename to `popper.css` though as I think there are some animations in there that are used by components that aren't popper (fade animations used in dialogs and scroll area)?

packages/radix-ui-themes/src/components/index.css
```diff
@@ -0,0 +1,54 @@
...
+@import './switch.css';
+@import './tab-nav.css';
+@import './tabs.css';
+@import './table.css';
+@import './text-area.css';
+@import './text-field.css';
+@import './text.css';
+@import './tooltip.css';
+
+.radix-themes:where([data-is-root-theme='true']) {
```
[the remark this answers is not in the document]
vladmoroz: `.radix-themes:where([data-is-root-theme='true'])` isn't really resetting anything, to me it feels like it is more of a generic component style that affects all portals, not really resetting

And then `reset.css` can become the base for that Reset component, given that all the other styles in this file are scoped to `.rt-reset` class

packages/radix-ui-themes/src/components/popper.css
vladmoroz: I'm cool with calling it `animations.css` under components (and/or even splitting it into `animations.css` and `popper.css`), what do you think?

benoitgrelard: I think we can leave it called `animations.css` personally.

benoitgrelard: We can split if this ever grows.

PR #266, opened 2024-01-29
RC release CSS fixes

vladmoroz:
Did an RC to test and found these issues

### DIFF
--- a/packages/radix-ui-themes/src/components/code.css
+++ b/packages/radix-ui-themes/src/components/code.css
@@ -123,10 +123,6 @@
         filter: var(--base-button-solid-high-contrast-hover-filter);
       }
     }
-    &:where(:focus-visible) .rt-Code {
-      outline: 2px solid var(--color-focus-root);
-      outline-offset: 2px;
-    }
   }
 }
 
@@ -147,11 +143,6 @@
       &:where(:hover) {
         background-color: var(--accent-a4);
       }
-    }
-    &:where(:focus-visible) .rt-Code {
-      /* Use gray outline when component color is gray */
-      outline: 2px solid var(--accent-8);
-      outline-offset: -1px;
     }
   }
 }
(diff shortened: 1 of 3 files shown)


packages/radix-ui-themes/src/components/code.css
```diff
@@ -123,10 +123,6 @@
         filter: var(--base-button-solid-high-contrast-hover-filter);
       }
     }
-    &:where(:focus-visible) .rt-Code {
-      outline: 2px solid var(--color-focus-root);
-      outline-offset: 2px;
-    }
```
vladmoroz, line 129: Stylelint didn't like the resulting selector (`:where(.rt-Link) .rt-Code:where(.rt-variant-solid):where(:focus-visible) .rt-Code` which also doesn't make sense).Wasn't really doing anything, probably a permutation of some original style that is not needed anymoroe. Did not hurt to remove it.

packages/radix-ui-themes/src/components/radio-card-group.css
```diff
@@ -94,8 +94,10 @@
     background-color: var(--color-surface);
     box-shadow: inset 0 0 0 1px var(--gray-a5);
 
-    &:where(:not([data-state='checked']):hover) {
-      box-shadow: inset 0 0 0 1px var(--gray-a7);
+    @media (hover: hover) {
+      &:where(:not([data-state='checked']):hover) {
+        box-shadow: inset 0 0 0 1px var(--gray-a7);
+      }
```
vladmoroz, line 100: `postcss-hover-media-feature` in WorkDS attempted to add `@media (hover: hover)`, but with complex selectors like this it tends to get into some endless loop crashing the node servers

packages/radix-ui-themes/src/helpers/props/text-align.prop.ts
```diff
@@ -4,7 +4,7 @@ const alignValues = ['left', 'center', 'right'] as const;
 
 const alignProp = {
   type: 'enum',
-  className: 'rt-r-align',
+  className: 'rt-r-ta',
```
vladmoroz, line 7: Found another broken classname
Looked through other props and haven't found anything suspicious, hope this one is the last

benoitgrelard (approved):

PR #272, opened 2024-01-30
Add CheckboxCardGroup

benoitgrelard:
https://github.com/radix-ui/themes/assets/1539897/31d5d058-5450-4860-8eaa-a560ecbee1b6


benoitgrelard:
@andy-hook adding you to this one because it has a bunch of logic in it which is more akin to a primitive.

packages/radix-ui-themes/src/components/radio-card-group.tsx
```diff
@@ -23,7 +23,7 @@ const RadioCardGroupRoot = React.forwardRef<RadioCardGroupRootElement, RadioCard
       marginPropDefs
     );
     return (
-      <Grid asChild>
+      <Grid asChild columns={props.columns ?? 'repeat(auto-fit, minmax(160px, 1fr))'}>
```
vladmoroz, line 26: Should we do this through the prop defs?

packages/radix-ui-themes/src/components/checkbox-card-group.css
```diff
@@ -0,0 +1,158 @@
...
+      }
+    }
+
+    &:where(:has(:disabled)) > * {
+      color: var(--gray-a9);
+    }
+
+    &:has(:focus-visible)::before {
+      outline: 2px solid var(--color-focus-root);
+      outline-offset: 2px;
```
vladmoroz, line 116: Don't think we should make it look like the Radio Card when it can look like Card

```suggestion
      outline-offset: -1px;
```

packages/radix-ui-themes/src/components/checkbox-card-group.css
```diff
@@ -0,0 +1,158 @@
...
+      box-shadow: 0 0 0 1px var(--gray-a5), var(--shadow-2);
+    }
+
+    &:where(:has(:disabled)) > * {
+      color: var(--gray-a9);
+    }
+
+    &:has(:focus-visible)::before {
+      outline: 2px solid var(--color-focus-root);
+      outline-offset: 3px;
```
vladmoroz, line 148:
```suggestion
      outline-offset: -1px;
```

packages/radix-ui-themes/src/components/checkbox-card-group.css
benoitgrelard: Ah yeah good point, that treatment is only really needed because of the checked state on radio cards.

packages/radix-ui-themes/src/components/radio-card-group.css
```diff
@@ -6,15 +6,18 @@
...
-  flex-shrink: 0;
+.rt-RadioCardGroupItemInner {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+
+  & * {
+    /* Avoid unintentional drag interactions (e.g. on images) */
+    pointer-events: none;
+  }
```
vladmoroz, line 20: - Is there any way we could avoid having an inner element? This would make passing a `style` prop much more viable, especially when overriding the default layout.
  - I scouted the internet for Radio Card examples in the wild and truncation by default is probably not such a great idea, e.g. [Apple](https://www.apple.com/shop/buy-mac/mac-pro/rack). Should we leave it to consumers to compose with `<Text truncate>`? (Just need to make sure it works, e.g. via `& > * { min-width: 0 }` on the items)
- There's a regression on the layout with icons:
  - https://themes-playground-git-checkbox-card-group-modulz.vercel.app/test-radio-card
  - https://themes-playground-modulz.vercel.app/test-radio-card
- Turbo nit: it was `& > *` and not `& *` so that `pointer-events: none` doesn't pollute the styles of the inner elements for a nicer web inspector DX

packages/radix-ui-themes/src/components/radio-card-group.css
```diff
@@ -121,17 +129,22 @@
...
-        box-shadow: 0 0 0 1px var(--gray-a5), var(--shadow-2);
+    &:where(:not(:disabled):active) {
+      transition-duration: 40ms;
+      box-shadow: 0 0 0 1px var(--gray-a5), var(--shadow-2);
+    }
+
+    &:where(:disabled) {
+      &,
+      & > * {
+        color: var(--gray-a9);
```
vladmoroz, line 147: Why do we need `& > *` for text color?

packages/radix-ui-themes/src/components/radio-card-group.tsx
vladmoroz: In `radio-card-group.props.ts`

packages/radix-ui-themes/src/components/radio-card-group.css
vladmoroz: Also thinking that we should default to justifying the content to center by default as a more vanilla option

vladmoroz:
Can we make the entire card gray when disabled? (Gray A2)

image

And then the background of the checkbox within the card would be transparent, but the regular disabled checkbox would stay the same

packages/radix-ui-themes/src/components/checkbox-card-group.css
```diff
@@ -0,0 +1,158 @@
...
+      &:where(:not(:has(:disabled)):hover) {
+        transition-duration: 40ms;
+        box-shadow: var(--card-classic-hover-box-shadow);
+      }
+    }
+
+    &:where(:not(:has(:disabled)):active) {
+      transition-duration: 40ms;
+      box-shadow: 0 0 0 1px var(--gray-a5), var(--shadow-2);
+    }
```
vladmoroz, line 140: Somehow found the `:active` style really annoying with the classic checkbox card (compared to the regular classic card), let's just remove it here?

packages/radix-ui-themes/src/components/radio-card-group.tsx
benoitgrelard: But these don't exist there…

packages/radix-ui-themes/src/components/checkbox-card-group.css
benoitgrelard: Yeah I found the same too! I'll remove them it for those.

packages/radix-ui-themes/src/components/radio-card-group.css
benoitgrelard: Yeah happy to drop the default truncation inside the component itself. 

> Should we leave it to consumers to compose with <Text truncate>

I also agree with this, that's how it works already when users have their own custom layout inside.

> There's a regression on the layout with icons

That's not a regression, I actually removed the default layout accommodating for icons, etc as I thought we'd say we perhaps shouldn't do it and let users lay things out outside of a trivial text inside the card? I just forgot to update this page to remove the icons.

But let me know if you think we should actually keep the convenience of flex + gap for text/icons by default.

packages/radix-ui-themes/src/components/checkbox-card-group.css
```diff
@@ -0,0 +1,163 @@
...
+  align-items: center;
+  gap: var(--checkbox-card-group-item-gap);
+  padding: var(--checkbox-card-group-item-padding);
+  border-radius: var(--checkbox-card-group-item-radius);
+
+  /* cursors */
+  cursor: var(--cursor-button);
+  &:where(:has(:disabled)) {
+    cursor: var(--cursor-disabled);
+  }
```
benoitgrelard, line 23: This isn't working for some reason, because of how the CSS gets reordered.
It's making me wonder if we have the same issue in the similar places where you've done that in your last PR?

packages/radix-ui-themes/src/components/checkbox-card-group.css
```diff
@@ -0,0 +1,163 @@
...
+}
+
+.rt-CheckboxCardGroupItemInner {
```
benoitgrelard, line 26: I've kept the inner for the checkbox rather than going with position absolute because this way it takes part into the layout, thus not reducing the amount of space available for text.

packages/radix-ui-themes/src/components/checkbox-card-group.tsx
```diff
@@ -0,0 +1,169 @@
...
+  required: boolean;
+  value?: string[];
+  onItemCheck(value: string): void;
+  onItemUncheck(value: string): void;
+};
+const CheckboxCardGroupContext = React.createContext<CheckboxCardGroupContextValue>({
+  disabled: false,
+  required: false,
+  onItemCheck() {},
+  onItemUncheck() {},
```
andy-hook, line 28: I wonder if it's worth using the context package to surface the composition error when the root hasn't been provided? alternatively, if we don't want to faff with the scopes we could guard and throw in a `useContext` hook of our own.

Not sure how important this is, though is does mean the DX differs from some of the other primitives.

andy-hook (approved):
Just a question on the context usage but otherwise this looks great to me. I'll defer to @vladmoroz on the styling side of things.

packages/radix-ui-themes/src/components/checkbox-card-group.css
vladmoroz: Hm, I don't follow how it helps; let's catch up about this

packages/radix-ui-themes/src/components/checkbox-card-group.tsx
benoitgrelard: oh yeah good point about the context scoping too, that's normally done in the primitive, so should do it here for now.

packages/radix-ui-themes/src/components/checkbox-card-group.css
```diff
@@ -0,0 +1,168 @@
...
+  position: relative;
+  /* pseudo element for focus outline */
+  &::before {
+    content: '';
+    display: block;
+    position: absolute;
+    inset: 0;
+    border-radius: inherit;
+    pointer-events: none;
+  }
```
vladmoroz, line 108: This is not used?

benoitgrelard: Good catch 🙂

packages/radix-ui-themes/src/components/checkbox-card-group.css
```diff
@@ -0,0 +1,168 @@
...
+.rt-CheckboxCardGroupItem {
+  display: flex;
+  align-items: center;
+  gap: var(--checkbox-card-group-item-gap);
+  box-sizing: border-box;
+  padding: var(--checkbox-card-group-item-padding-y) var(--checkbox-card-group-item-padding-x);
+  padding-right: calc(
+    var(--checkbox-card-group-item-padding-x) + var(--checkbox-card-group-item-checkbox-size) +
+      var(--checkbox-card-group-item-gap)
+  );
```
vladmoroz, line 19: I'd do `calc(var(--checkbox-card-group-item-padding-x) * 2 + var(--checkbox-card-group-item-checkbox-size));` to keep the checkbox looking a bit more natural with equal whitespace on the left and on the right around it. It feels more pinned to the right than part of the flex with the item.

image

packages/radix-ui-themes/src/components/radio-card-group.css
```diff
@@ -118,27 +126,28 @@
...
+      color: var(--gray-a9);
+    }
+
     &:where([data-state='checked']) {
       outline: 2px solid var(--accent-9);
       outline-offset: -1px;
+
+      &:where(:disabled) {
+        outline-color: var(--gray-8);
+      }
```
vladmoroz, line 150: We need to make sure this also works for high contrast

image

packages/radix-ui-themes/src/components/checkbox-card-group.tsx
```diff
@@ -0,0 +1,122 @@
...
+        {...checkboxGroupScope}
+        {...props}
+        ref={forwardedRef}
+        className={classNames(
+          'rt-reset',
+          'rt-BaseCheckboxRoot',
+          checkboxClassName,
+          'rt-CheckboxCardGroupItemCheckbox'
+        )}
+      >
```
vladmoroz, line 94: Should we pass the high contrast class manually here?

vladmoroz: To render a black checkbox

packages/radix-ui-themes/src/components/checkbox-card-group.css
benoitgrelard: Yes, that's better, thanks!

packages/radix-ui-themes/src/components/radio-card-group.css
benoitgrelard: Yep

packages/radix-ui-themes/src/components/checkbox-card-group.tsx
benoitgrelard: Oh nice catch, yeah they don't look right.

vladmoroz (approved):

PR #273, opened 2024-01-31
`asChild` prop

vladmoroz:
A pass on the availability of the `asChild` prop.

Generally:
- Most complex components like Checkbox, Switch, Slider, Tooltip, etc. that compose multiple primitive parts internally won't have `asChild`
- All or almost all panels (Card, Dialog Content, Popover Content, etc) should have `asChild`
- All components that can be used for layout purposes (Section, Container, Inset, Scroll Area) should have `asChild`
- All formatting components like Em, Strong, Code should have `asChild`

Also:
- All triggers that have `asChild={true}` by default will now throw a loud error if you don't pass a React Element as a child
- `as` and `asChild` are now in the prop defs so they can show up in the docs automatically


packages/radix-ui-themes/src/helpers/require-react-element.ts
```diff
@@ -0,0 +1,20 @@
...
+    throw Error(
+      `Expected a single React Element child, but got: ${React.Children.toArray(children)
+        .map((child) =>
+          typeof child === 'object' && 'type' in child && typeof child.type === 'string'
+            ? child.type
+            : typeof child
+        )
+        .join(', ')}`
+    );
+  }
```
vladmoroz, line 17: Would be nice to have this in the Slot primitive itself

packages/radix-ui-themes/src/helpers/get-root.tsx
```diff
@@ -0,0 +1,47 @@
...
+    let child = React.Children.only(children) as React.ReactElement;
+    const grandChildren = child.props.children;
+    return {
+      Root: ((props) => {
+        child = React.cloneElement(child, {
+          children: props.children,
+        });
+
+        // Make sure we don't pass `asChild` to DOM elements
+        const asChildProp = (Parent as unknown) === Slot ? undefined : asChild;
```
vladmoroz, line 31: @benoitgrelard I did see React complain about an unrecognized attribute

packages/radix-ui-themes/src/helpers/extract-props.ts
```diff
@@ -28,14 +28,25 @@ function extractProps<
...
 
-    const value = extractedProps[key];
+    // Apply the default value if the value is not a valid enum value
+    if (propDef.type === 'enum') {
+      const values = [propDef.default, ...propDef.values];
+
+      if (!values.includes(value) && !isResponsiveObject(value)) {
+        value = propDef.default;
+      }
+    }
```
vladmoroz, line 46: Fun, but something like `<Box as="whatever" />` actually rendered `<whatever className="rt-Box" />` because we were not enforcing that the prop value was what we expected

packages/radix-ui-themes/src/helpers/require-react-element.ts
benoitgrelard: I'm confused by this a little, it seems to replicate what `React.Children.Only` would do already? Don't you get this message (or very similar) anyway if you pass more than one child?

I've tried and if I pass nothing in a dialog trigger it fails silently, and if I pass 2 buttons, I still get the error from `React.Children.Only`, so when am I supposed to see this?

packages/radix-ui-themes/src/components/alert-dialog.tsx
```diff
@@ -46,6 +50,7 @@ const AlertDialogContent = React.forwardRef<AlertDialogContentElement, AlertDial
               <div className="rt-BaseDialogScrollPadding rt-AlertDialogScrollPadding">
                 <AlertDialogPrimitive.Content
                   {...contentProps}
+                  asChild={false}
```
benoitgrelard, line 53: Isn't this always going to override `asChild` from `contentProps`?
Also, I saw your note about panels like those needing `asChild`, what is the reasoning?

packages/radix-ui-themes/src/components/avatar.tsx
```diff
@@ -4,61 +4,86 @@ import * as React from 'react';
...
+  });
+
   return (
     // TODO as a rule, should we rather spread the props on root?
-    <AvatarPrimitive.Root
+    <AvatarRoot
       data-accent-color={color}
       data-radius={radius}
-      className={classNames('rt-AvatarRoot', className)}
+      className={classNames('rt-reset rt-AvatarRoot', className)}
```
benoitgrelard, line 34: nit:
```suggestion
      className={classNames('rt-reset', 'rt-AvatarRoot', className)}
```

packages/radix-ui-themes/src/components/badge.tsx
```diff
@@ -9,18 +10,19 @@ type BadgeElement = React.ElementRef<'span'>;
...
+  const Comp = asChild ? Slot : 'span';
   return (
-    <span
+    <Comp
       data-accent-color={color}
       data-radius={radius}
       {...badgeProps}
       ref={forwardedRef}
-      className={classNames('rt-Badge', className)}
+      className={classNames('rt-reset rt-Badge', className)}
```
benoitgrelard, line 25: nit:
```suggestion
      className={classNames('rt-reset', 'rt-Badge', className)}
```

packages/radix-ui-themes/src/components/dialog.tsx
```diff
@@ -45,6 +45,7 @@ const DialogContent = React.forwardRef<DialogContentElement, DialogContentProps>
               <div className="rt-BaseDialogScrollPadding rt-DialogScrollPadding">
                 <DialogPrimitive.Content
                   {...contentProps}
+                  asChild={false}
```
benoitgrelard, line 48: Same question as `AlertDialog`

packages/radix-ui-themes/src/components/callout.tsx
```diff
@@ -92,6 +95,7 @@ const CalloutText = React.forwardRef<CalloutTextElement, CalloutTextProps>(
         color={color}
         highContrast={highContrast}
         {...props}
+        asChild={false}
```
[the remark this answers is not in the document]
benoitgrelard: Oh I think I've realised why, it's because it's already removed by `ExtractPropsForTag` right?

That's not super obvious when reviewing, I wonder if we could do something about that? Rename the type helper maybe?

packages/radix-ui-themes/src/components/popover.tsx
```diff
@@ -44,6 +48,7 @@ const PopoverContent = React.forwardRef<PopoverContentElement, PopoverContentPro
             sideOffset={8}
             collisionPadding={10}
             {...contentProps}
+            asChild={false}
```
benoitgrelard, line 51: Same here

packages/radix-ui-themes/src/components/quote.tsx
```diff
@@ -1,13 +1,17 @@
...
 type QuoteElement = React.ElementRef<'q'>;
-interface QuoteProps extends PropsWithoutRefOrColor<'q'> {}
-const Quote = React.forwardRef<QuoteElement, QuoteProps>((props, forwardedRef) => (
-  <q {...props} ref={forwardedRef} className={classNames('rt-Quote', props.className)} />
-));
+type QuoteOwnProps = GetPropDefTypes<typeof emPropDefs>;
+interface QuoteProps extends PropsWithoutRefOrColor<'q'>, QuoteOwnProps {}
+const Quote = React.forwardRef<QuoteElement, QuoteProps>(({ asChild, ...props }, forwardedRef) => {
+  const Comp = asChild ? Slot : 'q';
+  return <Comp {...props} ref={forwardedRef} className={classNames('rt-Quote', props.className)} />;
```
benoitgrelard, line 13: Same about the `props.className` here

packages/radix-ui-themes/src/components/reset.css
```diff
@@ -3,12 +3,16 @@
...
 
 .rt-reset:where(a) {
-  cursor: var(--cursor-link);
+  cursor: auto;
   text-decoration: none;
   color: inherit;
   outline: none;
 }
 
+.rt-reset:where(a[href]) {
```
benoitgrelard, line 12: nice 👍

packages/radix-ui-themes/src/components/strong.tsx
```diff
@@ -1,13 +1,21 @@
...
-const Strong = React.forwardRef<StrongElement, StrongProps>((props, forwardedRef) => (
-  <strong {...props} ref={forwardedRef} className={classNames('rt-Strong', props.className)} />
-));
+type StrongOwnProps = GetPropDefTypes<typeof strongPropDefs>;
+interface StrongProps extends PropsWithoutRefOrColor<'strong'>, StrongOwnProps {}
+const Strong = React.forwardRef<StrongElement, StrongProps>(
+  ({ asChild, ...props }, forwardedRef) => {
+    const Comp = asChild ? Slot : 'strong';
+    return (
+      <Comp {...props} ref={forwardedRef} className={classNames('rt-Strong', props.className)} />
```
benoitgrelard, line 15: Same here for `props.className`

packages/radix-ui-themes/src/components/tabs.props.ts
```diff
@@ -1 +1,10 @@
+import { asChildProp } from '../helpers';
+
+const tabsContentPropDefs = {
+  asChild: asChildProp,
+} satisfies {
+  asChild: typeof asChildProp;
+};
+
+export { tabsContentPropDefs };
 export { baseTabListPropDefs as tabsListPropDefs } from './base-tab-list.props';
```
benoitgrelard, line 10: nit:

```suggestion
export { baseTabListPropDefs as tabsListPropDefs } from './base-tab-list.props';
export { tabsContentPropDefs };
```

packages/radix-ui-themes/src/components/tab-nav.props.ts
```diff
@@ -1 +1,10 @@
+import { asChildProp } from '../helpers';
+
+const tabNavLinkPropDefs = {
+  asChild: asChildProp,
+} satisfies {
+  asChild: typeof asChildProp;
+};
+
+export { tabNavLinkPropDefs };
 export { baseTabListPropDefs as tabNavPropDefs } from './base-tab-list.props';
```
benoitgrelard, line 10: nit:

```suggestion
export { baseTabListPropDefs as tabNavPropDefs } from './base-tab-list.props';
export { tabNavLinkPropDefs };
```

packages/radix-ui-themes/src/helpers/get-root.tsx
benoitgrelard: I still see one on the sink page but might be coming from somewhere?

benoitgrelard (commented):
Would be good to catchup on that one, I feel like there are some confusing things/potential mistakes? Would love to understand more behind the choice rationale.

Looks like there's an issue with callout which ends up spreading `asChild` to the DOM:
 
CleanShot 2024-02-01 at 12 19 45@2x

packages/radix-ui-themes/src/components/alert-dialog.tsx
vladmoroz: > Isn't this always going to override `asChild` from `contentProps`?

This is a mistake, I'll fix. I went back and forth on these panels a bit (alert dialog content, dialog content, hover card content, popover content) and forgot to remove `asChild={false}` after the initial iteration.

My reasoning here is that the modal panel layouts might need to be quite flexible, and since they don't come with any fancy interactions or semantics, there's no good reason we'd prevent you from asChild'ing them into other elements.

packages/radix-ui-themes/src/components/base-button.tsx
```diff
@@ -27,7 +27,7 @@ const BaseButton = React.forwardRef<BaseButtonElement, BaseButtonProps>((props,
   const {
     className,
     children,
-    asChild = false,
+    asChild,
```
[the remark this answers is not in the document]
vladmoroz: Good spot!

packages/radix-ui-themes/src/components/base-menu.props.ts
```diff
@@ -28,24 +28,30 @@ const baseMenuContentPropDefs = {
 };
 
 const baseMenuItemPropDefs = {
+  asChild: asChildProp,
```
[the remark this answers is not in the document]
vladmoroz: tbh I don't think we should allow that, but we already had some code in place to support it. Let's mull it over in person.

packages/radix-ui-themes/src/components/callout.tsx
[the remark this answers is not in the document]
vladmoroz: > Oh I think I've realised why, it's because it's already removed by `ExtractPropsForTag` right?

Yep. Do you have a better name in mind?

packages/radix-ui-themes/src/components/em.tsx
```diff
@@ -1,13 +1,17 @@
...
 type EmElement = React.ElementRef<'em'>;
-interface EmProps extends PropsWithoutRefOrColor<'em'> {}
-const Em = React.forwardRef<EmElement, EmProps>((props, forwardedRef) => (
-  <em {...props} ref={forwardedRef} className={classNames('rt-Em', props.className)} />
-));
+type EmOwnProps = GetPropDefTypes<typeof emPropDefs>;
+interface EmProps extends PropsWithoutRefOrColor<'em'>, EmOwnProps {}
+const Em = React.forwardRef<EmElement, EmProps>(({ asChild, ...props }, forwardedRef) => {
+  const Comp = asChild ? Slot : 'em';
+  return <Comp {...props} ref={forwardedRef} className={classNames('rt-Em', props.className)} />;
```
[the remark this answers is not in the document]
vladmoroz: Yep this is an oversight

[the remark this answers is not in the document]
vladmoroz: Did a pass in 3c7ef20be9d16a6151b9bcbdb03d4fcc24b23e00

packages/radix-ui-themes/src/helpers/get-root.tsx
vladmoroz: Found it, was extracting it from the wrong place
2c6eb447b6e4d166fb8aeb7d45102da42ca46507

packages/radix-ui-themes/src/helpers/require-react-element.ts
vladmoroz: This isn't really about ensuring that a single child is passed, more about ensuring an element is passed as a child, and not string / number / null / etc

Pedro and iirc someone else complained that it's unexpected how triggers like `<DialogTrigger>Open</DialogTrigger>` don't work and also don't throw any errors, so it looks just like a silent bug if you expected it to render a button like the primitives do

benoitgrelard (approved):

PR #281, opened 2024-02-05
Improve ESM compatibility

vladmoroz:
- Improve ESM compatibility
- Use esbuild for compiling TypeScript
  - Technically we could keep using `tsc`, but I evaluated esbuild as part of the whole process and decided to keep it because it builds the project ~3x faster, which feels much more responsive in development
  - `tsc` is still used to emit declarations. Changed to use the `incremental` flag in development for better performance.
- Bump build target to `ES2020` (from `ES2018`)

***

Built package comparison:
https://www.npmjs.com/package/@radix-ui/themes/v/3.0.0-rc.1?activeTab=code
https://www.npmjs.com/package/@radix-ui/themes/v/2.0.3?activeTab=code

***

[publint](https://publint.dev/), now:

image

publint, before:

image

***

Astro (ESM):

image

Remix (ESM):

image

NextJS AuthKit (using `branded-ui` package compiled from the CJS version of Radix Themes):

image

NextJS WorkOS dashboard (using `design-system` package compiled from the CJS version of Radix Themes):

image


packages/radix-ui-themes/esbuild-cjs.mjs
```diff
@@ -0,0 +1,22 @@
+import esbuild from 'esbuild';
+
+const dir = 'dist/cjs';
+
+const options = {
+  entryPoints: ['src/**/*.ts*'],
+  outdir: dir,
+  format: 'cjs',
+  target: 'es2020',
+  sourcemap: true,
```
vladmoroz, line 10: @benoitgrelard do we want to keep including source maps in the npm package?

See these issues:
https://github.com/radix-ui/themes/issues?q=is%3Aissue+is%3Aopen+source+map

What we have right now doesn't make sense to me:
- Source maps point to the original TypeScript files
- We ship source maps in the package
- Webpack complains that source maps point to files that don't exist

I think we should either drop them from non-dev builds or include the entire `src` contents into the npm package

benoitgrelard: I think sourcemaps are good to have. We should keep them and publish the `src` on npm.

packages/radix-ui-themes/src/components/callout.parts.ts
```diff
@@ -1,3 +1,3 @@
-export { CalloutRoot as Root } from './callout';
-export { CalloutIcon as Icon } from './callout';
-export { CalloutText as Text } from './callout';
+export { CalloutRoot as Root } from './callout.js';
+export { CalloutIcon as Icon } from './callout.js';
+export { CalloutText as Text } from './callout.js';
```
benoitgrelard, line 3: Unrelated as it came from the another PR, but is there a reason why those `.parts` files can't be combined in a single export call?

ie. for this use-case:

```ts
export {
  CalloutRoot as Root,
  CalloutIcon as Icon,
  CalloutText as Text
} from './callout.js';
```

packages/radix-ui-themes/src/components/flex.props.ts
```diff
@@ -1,5 +1,5 @@
-import { asChildProp, gapProps } from '../helpers';
-import type { PropDef } from '../helpers';
+import { asChildProp, gapProps } from '../helpers/index.js';
+import type { PropDef } from '../helpers/index.js';
```
benoitgrelard, line 2: This is so odd to read haha, a type import from a JS file 😀

packages/radix-ui-themes/src/components/callout.parts.ts
vladmoroz: @penx do you happen to know whether the export style would have any effect?

packages/radix-ui-themes/src/components/flex.props.ts
vladmoroz: I know, breaks my brain too

packages/radix-ui-themes/esbuild-cjs.mjs
vladmoroz: Added

benoitgrelard (approved):

packages/radix-ui-themes/src/components/callout.parts.ts
penx: @vladmoroz sorry for the late reply, no this makes sense to change it. I think I just did it that way as it was faster the way I was doing the same copy/paste/type over several files with multi line cursors.

Issue #6, opened 2023-08-09
[ThemePanel - Bug] Appearance switch of ThemePanel doesn't work properly


w1that:
I tried the ThemePanel at [play ground](https://www.radix-ui.com/themes/playground) and wanted to add it to my React application. 

I use `<Theme appearance="dark" accentColor="tomato" grayColor="sage">`, so I set the appearance to `dark` as default. When I start the app, the Light radio input is selected, while the Panel is in dark theme. After clicking on Light radio input, it turns to Light theme but the dark theme radio input can't be selected. 

I guess it is because of class of the `ThemePanel` can't find a way to change from `light` to `dark`

shfrmn:
I also encountered this bug yesterday, but didn't want to open the first issue in the new repo 😅 

@w1that As an ugly temporary solution you can add this to your root component.
```jsx
  useEffect(() => {
    const root = document.getElementById("root")!
    const script = root.querySelector("script")!
    const scriptCopy = document.createElement("script")
    scriptCopy.innerText = script.innerText!
    root.appendChild(scriptCopy!)
  }, [])
```

The bug is caused by the fact `ExplicitRootAppearanceScript` doesn't actually execute or misfires in some way. I wasn't able to immediately understand what happens.
https://github.com/radix-ui/themes/blob/39f5a438db0d8496134d4bc35d71899fee1b2562/packages/radix-ui-themes/src/theme.tsx#L84

vladmoroz:
Thanks for raising the issue, we'll look into it. What is your setup exactly?

If you need just the dark mode, could you try this:
1. If you have access to the `html` element, set `class="dark"` or `class="dark-theme"`
2. Then, just don't pass the `appearance` prop on the `Theme` component

shfrmn:
@vladmoroz Should be very easy to reproduce. Just create a new react app using yarn+vite+typescript, then yarn add the library and wrap everything with `<Theme appearance="dark">`

Regarding adding `class="dark"` no doubt it will start working after that. What I found is that the code inside `ExplicitRootAppearanceScript` works perfectly when executed manually in the console, so I figured it probably just doesn't run when you inject the script tag for whatever reason. In the solution I posted above I just reinsert the script tag as you can see and the theme setup happens as you intended.

w1that:
As @shfrmn said, I just created a new React app with CRA-typescript template. 

I use the `Template` and `TemplatePanel` in index.tsx like this (_as suggested at the docs_):
```
[...]

Issue #280, opened 2024-02-05
[Theme Appearance] The appearance is always `light` in Next.js `not-found` and `error` pages


simonfarah:
I have a Next.js 14 project in development, and I am having this issue. I have my Radix Theme provider setup correctly in the root layout with a dark appearance and everything's working great. However, in the Next.js `not-found` and `error` pages, the appearance is always light unless I manually add again the provider to those pages, and end up having two radix providers on top of each other. I noticed the `dark` class name disappear from the html element, and that's maybe due to how Next.js behave in those routes or maybe a Radix Theme bug?

vladmoroz:
This is probably a Next.js quirk. I vaguely remember that we ran into a similar issue in one of our projects, but can’t remember the exact resolution because we've changed more to how theming works there after. Some ideas I'd suggest, assuming you are using dark mode only:
- Remove `appearance="dark"` from the `Theme` provider and manually set `<html className="dark" style={{ color-scheme: "dark" }}>` in your root layout.
- Or, if the workaround with nested Theme providers works, continue using that—it’s fine technically, they are designed to be nested.

Issue #168, opened 2023-11-30
Cant override --color-background token


Hackjamin:
Hey Folks,

i tried to override the --color-background in a tokens.css of my next 13 / app router project. But it seems like it is resetting all the time.

The light mode should have a "beige" background, but the rendered html file still contains this:
```html
<style>
            :root, .light, .light-theme {
                --color-page-background: white;
            }

            .dark, .dark-theme {
                --color-page-background: var(--sage-1);
            }

            body {
                background-color: var(--color-page-background);
            }
        </style>
``` 

layout.tsx
```javascript
import '@radix-ui/themes/styles.css'
import type { Metadata } from 'next'
import { Theme, ThemePanel } from '@radix-ui/themes'
import Footer from '@/components/Footer'
import './globals.scss'
import './tokens.css'
``` 

tokens.css
```css
@font-face {
  font-family: Montserrat;
  src: local('/fonts/Montserrat-Italic-VariableFont_wght.ttf');
}
.radix-themes {
  --default-font-family: 'Montserrat';
  --heading-font-family: 'Montserrat';
  --quote-font-family: 'Montserrat';
  --font-weight-light: 300;
  --font-weight-regular: 400;
  --font-weight-medium: 500;
  --font-weight-bold: 700;
}

:where(.radix-themes) {
  --color-background: beige;
  --color-overlay: var(--black-a6);
  --color-panel-solid: beige;
  --color-panel-translucent: rgba(255, 255, 255, 0.8);
  --color-surface: rgba(255, 255, 255, 0.9);
}
:is(.dark, .dark-theme),
:is(.dark, .dark-theme) :where(.radix-themes:not(.light, .light-theme)) {
  --color-background: var(--gray-1);
  --color-overlay: var(--black-a8);
  --color-panel-solid: var(--gray-2);
  --color-panel-translucent: var(--gray-2-translucent);
  --color-surface: rgba(0, 0, 0, 0.25);
}
``` 

or did i miss something here?

ajdiyassin:
I think you should move your color tokens inside
.radix-themes{}

ajdiyassin:
I fixed this issue using Layers https://developer.mozilla.org/en-US/docs/Web/CSS/@layer

```
@layer tw_base, mantine, radix_ui, tw_components, tw_utilities;
@import 'tailwindcss/base' layer(tw_base);
@import 'tailwindcss/components' layer(tw_components);
@import 'tailwindcss/utilities' layer(tw_utilities);
[...]

Issue #111, opened 2023-10-08
Can't switch theme appearance


Asuka109:
You can reproduce it by:

1. visit the demo: https://stackblitz.com/edit/stackblitz-starters-ezzfyb?file=src%2FApp.tsx
2. switch appearance to dark by theme panel.

If you print the value of appearance, you will find that it becomes `dark` and then quickly changes back to `light`. It seems that this is because the `Theme` component failed to add the `dark` class name to `documentElement`.

I think this is related to the implementation of the Theme.

https://github.com/radix-ui/themes/blob/392d620afe8cd6cc4489ed56e4289ca05c49986e/packages/radix-ui-themes/src/theme.tsx#L83-L93

The script injected by `innerHTML` won't be executed. Run the script via `useEffect()` can resolve the issue.

https://github.com/radix-ui/themes/blob/392d620afe8cd6cc4489ed56e4289ca05c49986e/packages/radix-ui-themes/src/theme.tsx#L250-L265

Or adding some fallbacks here might be helpful.

Asuka109:
If this is unexpected, I am willing to submit a pull request to fix it.

libantheworld:
I had a similar issue and it turned out to be related to hydration, I was setting a new date with useState() and updating it within a UseEffect but by the time it build it wasn't matching. It might help to look for a similar bug.

elmdecoste:
I'm having the same problem, are there any updates here?

nomyfan:
Same here

claudiug:
same

mattsrobot:
I'm also having users having problems selecting the theme, and I haven't been able to reproduce it yet.

May I also is there a specific browser / os this affects?

robhadfield:
The theme panel works fine for me within my app locally but I did have issues with the theme not re-rendering when dynamically changing it within the `<Theme appearance={theme}>` properties.

I fixed this by toggling classes `dark` and `light` on the body tag.

```
// Theme name "dark" or "light" as strings coming from the store...
const theme = useYourStore((store) => store.theme); // I use zustand
```

Then simply adding it to the body:
```
<body className={theme}>
  {children}
</body>
```

Real-Music:
Same here

PR #282, opened 2024-02-05
Simplify dark theme logic

vladmoroz:
- Set `min-height: 100vh` on the root `Theme` component
- Fix an issue when in certain situations `hasBackground` prop value would have no effect
- Refine the logic for when `Theme` has a background color by default.
  - When there is no explicit `hasBackground` prop value is passed, `Theme` will display background color:
    - When it is the root `Theme` component
    - When it has an explicit appearance value, e.g. `<Theme apperance="light">` or `<Theme apperance="dark">`
  - Body background color and is no longer set automatically. The background color is now provided by the root `Theme` by default.
    - [**Breaking**] The CSS variable `--color-page-background` is no longer available. In most cases, it can be safely replaced with `--color-background` available on the `.radix-themes` element.
    - `suppressHydrationWarning` on `html` is no longer needed (unless required by other libraries, like `next-themes`)

Closes https://github.com/radix-ui/themes/issues/6, https://github.com/radix-ui/themes/issues/19, https://github.com/radix-ui/themes/issues/22, https://github.com/radix-ui/themes/issues/26, https://github.com/radix-ui/themes/issues/111, https://github.com/radix-ui/themes/issues/168, https://github.com/radix-ui/themes/issues/280

### DIFF
--- a/apps/playground/app/test-theme-toggle/page.tsx
+++ b/apps/playground/app/test-theme-toggle/page.tsx
@@ -1,0 +1,82 @@
+'use client';
+
+import * as React from 'react';
+import { Theme, Flex, ThemePanel, IconButton } from '@radix-ui/themes';
+import { useTheme } from 'next-themes';
+import { SunIcon, MoonIcon } from '@radix-ui/react-icons';
+import { NextThemeProvider } from '../next-theme-provider';
+
+export default function Test() {
+  return (
+    <html lang="en" suppressHydrationWarning>
+      <body>
+        <NextThemeProvider>
+          <Theme asChild>
+            <div id="root">
+              <ThemeTogglePanel />
+              <Flex height="100vh" width="100vw" align="center" justify="center">
+                <ThemeToggle />
+              </Flex>
+            </div>
+          </Theme>
+        </NextThemeProvider>
+      </body>
+    </html>
+  );
+}
+
+function ThemeTogglePanel() {
+  const { systemTheme, setTheme } = useTheme();
+  return (
+    <ThemePanel
+      onAppearanceChange={(newTheme) => {
+        const newThemeMatchesSystem = newTheme === systemTheme;
+        setTheme(newThemeMatchesSystem ? 'system' : (newTheme as 'light' | 'dark'));
+      }}
+    />
+  );
+}
+
+function ThemeToggle({ children, ...props }: React.ComponentPropsWithoutRef<typeof IconButton>) {
+  const { theme, systemTheme, setTheme } = useTheme();
+
+  return (
+    <>
+      <style>{`
+        :root, .light, .light-theme {
+          --theme-toggle-sun-icon-display: block;
+          --theme-toggle-moon-icon-display: none;
+        }
+        .dark, .dark-theme {
+          --theme-toggle-sun-icon-display: none;
+          --theme-toggle-moon-icon-display: block;
+        }
+      `}</style>
+
+      <IconButton
+        size="3"
+        variant="ghost"
+        color="gray"
+        onClick={() => {
+          // Set 'system' theme if the next theme matches the system theme
+          const resolvedTheme = theme === 'system' ? systemTheme : theme;
+          const newTheme = resolvedTheme === 'dark' ? 'light' : 'dark';
+          const newThemeMatchesSystem = newTheme === systemTheme;
+          setTheme(newThemeMatchesSystem ? 'system' : newTheme);
+        }}
+        {...props}
+      >
+        <SunIcon
+          width="16"
+          height="16"
+          style={{ display: 'var(--theme-toggle-sun-icon-display)' }}
+        />
+        <MoonIcon
+          width="16"
+          height="16"
+          style={{ display: 'var(--theme-toggle-moon-icon-display)' }}
+        />
+      </IconButton>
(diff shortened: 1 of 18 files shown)


packages/radix-ui-themes/src/theme.tsx
```diff
@@ -82,61 +82,27 @@ const ThemeRoot = React.forwardRef<ThemeImplElement, ThemeRootProps>((props, for
...
-        <style
-          dangerouslySetInnerHTML={{
-            __html: `
-:root, .light, .light-theme { --color-page-background: white; }
-.dark, .dark-theme { --color-page-background: var(--${resolvedGrayColor}-1); }
-body { background-color: var(--color-page-background); }
-`,
-          }}
-        />
-      )}
```
vladmoroz, line 117: Removing this is made possible by:
- The root `Theme` covers `100vh` and gets the background color, so we don't care about the body background anymore
- Since we don't care about the body background anymore, we don't need the theme classes on the html/body
- Now we don't need to inject the script and styles, suppress the html hydration warning, and to sync root appearance class in dev with the prop value

packages/radix-ui-themes/src/theme.tsx
```diff
@@ -176,12 +142,9 @@ const ThemeImpl = React.forwardRef<ThemeImplElement, ThemeImplProps>((props, for
   } = props;
   const Comp = asChild ? Slot : 'div';
   const resolvedGrayColor = grayColor === 'auto' ? getMatchingGrayColor(accentColor) : grayColor;
-  const isExplicitAppearance = props.appearance !== undefined && props.appearance !== 'inherit';
-  const isExplicitGrayColor = props.grayColor !== undefined;
-  const shouldHaveBackground =
-    !isRoot &&
-    (hasBackground === true ||
-      (hasBackground !== false && (isExplicitAppearance || isExplicitGrayColor)));
```
vladmoroz, line 184: We had some situations in AuthKit when passing an explicit `hasBackground` didn't have any effect

I also removed `isExplicitGrayColor` condition to simplify the logic as it was unintuitive. I recollect that the argument was to set background color when the gray tint is changed so that is visible in dark mode, but imo it's too fringe vs the complexity we are getting.

packages/radix-ui-themes/src/theme-panel.tsx
```diff
@@ -621,5 +638,30 @@ function upperFirst(string: string) {
...
+    root.classList.remove('light', 'dark');
+    root.style.colorScheme = appearance;
+    root.classList.add(appearance);
+  }
+
+  if (!hasLightTheme && !hasDarkTheme && !hasLight && !hasDark) {
+    root.style.colorScheme = appearance;
+    root.classList.add(appearance);
+  }
+}
```
vladmoroz, line 664: I went the other way around @benoitgrelard here...

Before, when using the appearance theme toggle in the theme panel when initial `appearance` was `inherit`, it was forced into `light`/`dark`.

Instead, I'm now preserving the original `appearance="inherit"` prop and toggling the class on the html element, and the prop value is changed only when an explicit `appearance` value is used.

apps/playground/app/test-theme-toggle/page.tsx
vladmoroz, line 1: This page re-creates what we have in the Radix docs where the `onAppearanceChange` need came up.

packages/radix-ui-themes/src/components/index.css
```diff
@@ -49,7 +49,10 @@
 @import './tooltip.css';
 
 .radix-themes:where([data-is-root-theme='true']) {
-  /* create new stacking context on the root `Theme` so layered components work out of the box */
+  /* Create a new stacking context on the root `Theme` so layered components work out of the box */
   position: relative;
   z-index: 0;
+
+  /* Make sure root `Theme` background covers the viewport */
+  min-height: 100vh;
```
vladmoroz, line 57: We had to add this style in every single project we used Radix Themes.

I think we shouldn't have the constraint of using Radix Themes in individual "sections" of an app; it's still going to be achievable with some friction, but the common case is simplified a ton.

packages/radix-ui-themes/src/theme-panel.css
```diff
@@ -32,7 +32,7 @@
 .rt-ThemePanelSwatchInput {
   outline-offset: 2px;
 
-  &:where(:checked) {
+  &:where([data-checked]) {
```
vladmoroz, line 35: I can't figure out why, but even on `main` the theme panel works weird on the /test-appearance page

CSS is not picking up `:checked` until there's a re-render 🤷

packages/radix-ui-themes/src/components/index.css
lucasmotta: Could this potentially be a breaking change for existing users?
And have you considered using [100dvh](https://web.dev/blog/viewport-units) instead?

vladmoroz: Yep, we quite a few sensitive changes have snowballed, so the next public release will be 3.0.0

Will add 100dvh, forgot about it 👍

benoitgrelard (approved):

ViktorQvarfordt:
These fixes sound great! When will they be released?

PR #285, opened 2024-02-05
Make sure `asChild` prop is not passed to DOM elements

vladmoroz:
When testing the new version in the WorkOS monorepo I noticed a warning about `asChild={undefined}` being passed to the Card’s DOM element

### DIFF
--- a/packages/radix-ui-themes/src/helpers/get-root.tsx
+++ b/packages/radix-ui-themes/src/helpers/get-root.tsx
@@ -8,7 +8,7 @@ interface GetChildrenArgs<T extends React.ElementType> {
 }
 
 export const getRoot = <
-  T extends React.ElementType,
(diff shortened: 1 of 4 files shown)


packages/radix-ui-themes/src/helpers/get-root.tsx
```diff
@@ -8,7 +8,7 @@ interface GetChildrenArgs<T extends React.ElementType> {
 }
 
 export const getRoot = <
-  T extends React.ElementType,
+  T extends React.ComponentType<any> | keyof JSX.IntrinsicElements,
```
vladmoroz, line 11: React.ElementType is different between workos and this repo because @types/react versions diverge, which our build toolchain complained about. This yields a similar result, but without the potential clashes of the type versions.

PR #286, opened 2024-02-05
Improve Radio Card focus states

vladmoroz:
Realised we had already solved this problem in the theme panel component

https://github.com/radix-ui/themes/assets/8441036/d1c253f3-de54-46ba-92e3-e0d14198e07d

### DIFF
--- a/packages/radix-ui-themes/changelog.md
+++ b/packages/radix-ui-themes/changelog.md
@@ -30,6 +30,11 @@
   - Ensure all elements that have padding or border styles use `box-sizing: border-box`
   - Ensure that disabled cursor styles are applied correctly
   - Tweak the background color of all `variant="soft"` menu items
+  - Replace `--color-focus-root`, `--color-selection-root`, `--color-autofill-root` with a focus color scale, e.g. `--focus-1` – `--focus-12`, and `--focus-a1` – `--focus-a12`.
+    - [**Breaking**] If you were overriding the above tokens or using them in your custom components, you’ll need to references the new color scale.
+      - `--color-autofill-root` is replaced by `--focus-a3`
+      - `--color-focus-root` is replaced by `--focus-8`
+      - `--color-selection-root` is replaced by `--focus-a5`
 - 8 new components
   - `CheckboxCardGroup`
   - `DataList`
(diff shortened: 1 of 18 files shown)


packages/radix-ui-themes/changelog.md
```diff
@@ -30,6 +30,11 @@
   - Ensure all elements that have padding or border styles use `box-sizing: border-box`
   - Ensure that disabled cursor styles are applied correctly
   - Tweak the background color of all `variant="soft"` menu items
+  - Replace `--color-focus-root`, `--color-selection-root`, `--color-autofill-root` with a focus color scale, e.g. `--focus-1` – `--focus-12`, and `--focus-a1` – `--focus-a12`.
```
benoitgrelard, line 33: This feels kinda weird, ie have selection/autofill use this new "focus" scale.
I wonder if we should rename it to something more generic like `--root-accent-xx`? It's not really anything to do with focus, but rather an accent scale which matches to root accent scale regardless of colour overrides (except for grays, but I think we can ignore that part).

vladmoroz: It does change based on the component accent, e.g. <Text color="red"> changes that color to red. This basically references the closest non-gray accent color for a given element.

I did ponder the naming, but ended up using `--focus-xx` because almost with any design you want autofill and focus colors to match in order to look good; and then the selection color follows the same rule as for which accent is used. E.g. when you have a focused input, it looks about as odd when its text selection color mismatches its focus outline.

benoitgrelard: > It does change based on the component accent, e.g. changes that color to red.

Oh yeah, that's true, it's the opposite of what I was thinking it was. Forgot about that!

> This basically references the closest non-gray accent color for a given element.

That's a good way to explain it yeah, I had it backwards.

> I did ponder the naming, but ended up using --focus-xx because almost with any design you want autofill and focus colors to match in order to look good; and then the selection color follows the same rule as for which accent is used. E.g. when you have a focused input, it looks about as odd when its text selection color mismatches its focus outline.

Fair then!

benoitgrelard (approved):

PR #290, opened 2024-02-07
Add `@radix-ui/themes/props` and `@radix-ui/themes/helpers` entry points

vladmoroz:
Rework the project structure and add `@radix-ui/themes/props` and `@radix-ui/themes/helpers` entry points to make it possible to build your own component library on top of Radix Themes using the same techniques.
- Note: you might need to use `"moduleResolution": "NodeNext"` with your compiler to access these paths
- Note: changes to prop defs and helpers won’t be covered by semver

Overview of the new structure:
- `src/components` is mostly unchanged. It still includes all the components.
  - `ThemePanel` and `Theme` were moved into this folder so that we don't treat them as much of a snowflake components anymore.
  - `icons.tsx` was also moved into components and exported to allow access to the caret icons and chevrons used for the Select and Dropdown Menu. However, the info icon was moved inline into `ThemePanel` as it's used just there.
- `src/helpers` includes everything that is not a component and not a prop definition.
- `src/props` includes just the prop definitions. Nothing is imported from helpers or components to avoid circular dependencies.
  - A bunch of prop-related helpers were relocated from props files into `src/helpers`, like `extractMarginProps` or the type helper to get component props without the `color` attribute
  - Theme props were refactored to mostly go into `theme.props.ts`
 
I also exposed the PostCSS plugins via "files" because we want to re-use them in the WorkOS design system

### DIFF
--- a/packages/radix-ui-themes/src/components/context-menu.tsx
+++ b/packages/radix-ui-themes/src/components/context-menu.tsx
@@ -17,7 +17,6 @@ import { ThickCheckIcon, ThickChevronRightIcon } from './icons.js';
 
 import type { ComponentPropsWithoutColor } from '../helpers/index.js';
 import type { GetPropDefTypes } from '../props/index.js';
-import { baseMenuContentPropDefs } from './base-menu.props.js';
 
 interface ContextMenuRootProps
   extends ComponentPropsWithoutColor<typeof ContextMenuPrimitive.Root> {}
@@ -57,7 +56,7 @@ const ContextMenuContent = React.forwardRef<ContextMenuContentElement, ContextMe
     } = props;
     const { className, children, color, container, forceMount, ...contentProps } = extractProps(
       props,
-      baseMenuContentPropDefs
+      contextMenuContentPropDefs
     );
     const resolvedColor = color || themeContext.accentColor;
     return (
@@ -298,7 +297,7 @@ const ContextMenuSubContent = React.forwardRef<
   const { size, variant, color, highContrast } = React.useContext(ContextMenuContentContext);
   const { className, children, container, forceMount, ...subContentProps } = extractProps(
     { size, variant, color, highContrast, ...props },
-    baseMenuContentPropDefs
+    contextMenuContentPropDefs
   );
   return (
     <ContextMenuPrimitive.Portal container={container} forceMount={forceMount}>
--- a/packages/radix-ui-themes/src/components/dropdown-menu.tsx
+++ b/packages/radix-ui-themes/src/components/dropdown-menu.tsx
@@ -17,7 +17,6 @@ import { ThickCheckIcon, ThickChevronRightIcon } from './icons.js';
 
 import type { ComponentPropsWithoutColor } from '../helpers/index.js';
 import type { GetPropDefTypes } from '../props/index.js';
-import { baseMenuContentPropDefs } from './base-menu.props.js';
 
 interface DropdownMenuRootProps
   extends ComponentPropsWithoutColor<typeof DropdownMenuPrimitive.Root> {}
@@ -300,7 +299,7 @@ const DropdownMenuSubContent = React.forwardRef<
   const { size, variant, color, highContrast } = React.useContext(DropdownMenuContentContext);
   const { className, children, container, forceMount, ...subContentProps } = extractProps(
     { size, variant, color, highContrast, ...props },
-    baseMenuContentPropDefs
+    dropdownMenuContentPropDefs
   );
   return (
     <DropdownMenuPrimitive.Portal container={container} forceMount={forceMount}>


packages/radix-ui-themes/src/props/prop-def.ts
```diff
@@ -73,4 +79,13 @@ type GetPropDefTypes<P> = {
   [K in keyof P]?: GetPropDefType<P[K]>;
 };
 
-export type { PropDef, ResponsivePropDef, GetPropDefTypes };
+export { breakpoints };
+export type {
+  PropDef,
+  GetPropDefTypes,
```
vladmoroz, line 85: I kind of feel like `GetPropDefTypes` is more of a helper, but didn't want to put it into helpers because we'd end up with helpers importing props and props importing helpers

packages/radix-ui-themes/changelog.md
```diff
@@ -5,6 +5,9 @@
 - General
   - Improve ESM compatibility
   - Improve tree-shaking of individual component parts
+  - [**Breaking**] Remove component prop definitions and internal helpers from the root `@radix-ui/themes` import entry point and export them from `@radix-ui/themes/props` and `@radix-ui/themes/helpers` to make it possible to build your own component library on top of Radix Themes using the same techniques.
+    - Note: you might need to use `"moduleResolution": "NodeNext"` with your compiler to access these paths
+    - Note: changes to prop defs and helpers won’t be covered by semver
```
benoitgrelard, line 10: I wonder if technically changes to prop defs are covered by semver as they are linked to changes in the component's APIs anyway? Or I guess you meant that if we somehow change the structured way those are exposed (not changing the props themselves) then we don't cover these changes via semver?

packages/radix-ui-themes/src/components/checkbox.tsx
```diff
@@ -4,15 +4,17 @@ import * as React from 'react';
 import classNames from 'classnames';
 import * as CheckboxPrimitive from '@radix-ui/react-checkbox';
 import { checkboxPropDefs } from './checkbox.props.js';
-import { extractProps, marginPropDefs } from '../helpers/index.js';
-import { ThickCheckIcon } from '../icons.js';
+import { extractProps } from '../helpers/index.js';
+import { marginPropDefs } from '../props/index.js';
```
benoitgrelard, line 8: Would it make sense to always combine imports from the barrel file (equivalent of `@radix-ui/themes/props` and `@radix-ui/themes/helpers`?

So here for example:

```tsx
import { checkboxPropDefs, marginPropDefs } from '../props/index.js';
```

packages/radix-ui-themes/src/components/context-menu.tsx
```diff
@@ -12,21 +12,23 @@ import {
...
 import { extractProps, requireReactElement } from '../helpers/index.js';
-import { Theme, useThemeContext } from '../theme.js';
-import { ThickCheckIcon, ThickChevronRightIcon } from '../icons.js';
+import { Theme, useThemeContext } from './theme.js';
+import { ThickCheckIcon, ThickChevronRightIcon } from './icons.js';
 
-import type { PropsWithoutRefOrColor, GetPropDefTypes } from '../helpers/index.js';
+import type { ComponentPropsWithoutColor } from '../helpers/index.js';
+import type { GetPropDefTypes } from '../props/index.js';
 import { baseMenuContentPropDefs } from './base-menu.props.js';
```
benoitgrelard, line 20: weird, we shouldn't need to import the base right as we re-export from it? Same with dropdown menu

packages/radix-ui-themes/changelog.md
vladmoroz: Yeah, like if we decide to remove/rework a prop def, e.g. something odd like `inheritedColorProp`, this shouldn't be covered by semver. Semver would of course still cover changes to the public component API, but the internal prop def changes should be up to us until this area is more mature and until we know about more similar use cases with a design system that extends Radix Themes.

packages/radix-ui-themes/src/components/checkbox.tsx
vladmoroz: I am slightly biased towards keeping the component portable with its props, but not a strong preference if you prefer cleaner imports. E.g. if we wouldn't have to export the prop defs for the docs, I wouldn't put them into prop index at all.

benoitgrelard: Let's keep as is for now then.

packages/radix-ui-themes/src/components/context-menu.tsx
vladmoroz: Oh right, updated

packages/radix-ui-themes/src/components/dropdown-menu.tsx
```diff
@@ -300,7 +299,7 @@ const DropdownMenuSubContent = React.forwardRef<
   const { size, variant, color, highContrast } = React.useContext(DropdownMenuContentContext);
   const { className, children, container, forceMount, ...subContentProps } = extractProps(
     { size, variant, color, highContrast, ...props },
-    baseMenuContentPropDefs
```
benoitgrelard, line 300: curious why we don't see 2 changes in that diff for dropdown menu too? Shouldn't there be one for the content part too?

vladmoroz: I think we were just random in this file, `baseMenuContentPropDefs` isn't used anywhere it shouldn't be anymore

image

benoitgrelard (approved):

PR #291, opened 2024-02-08
Fix cosmetic bugs from RC tests

vladmoroz:
- Make sure inline elements like Badge and Code that rely on padding for their height don't stretch vertically in Flex/Grid layouts, similarly to how fixed-height elements work
- Fix `a` reset regression when links would get an unintended `inline-block` style by default
- Rework Text Field and Text Area
  - Make sure that the font weight is not inherited, e.g. from a wrapping `<label>` element
  - Add styles for inputs of different types, like date and search
  - Got rid of the internal Chrome part given [good support](https://caniuse.com/css-has) for `:has`, which simplifies greatly our implementation and allows the `style` prop to work more predictably.
  - Older browser that don't support `:has` will render the disabled and autofilled styles as regular state; otherwise these components are still going to be fully functional.

### DIFF
--- a/packages/radix-ui-themes/src/components/text-field.css
+++ b/packages/radix-ui-themes/src/components/text-field.css
@@ -24,7 +24,9 @@
 }
 
 .rt-TextFieldInput {
+  /* Flex layout edge cases — make sure the input can grow, but won't break out of the container */
   flex-grow: 1;
+  min-width: 0;
 
   /* Fix date inputs alignment in Chrome and Safari */
   display: flex;
--- a/packages/radix-ui-themes/src/components/text-field.css
+++ b/packages/radix-ui-themes/src/components/text-field.css
@@ -1,36 +1,94 @@
 .rt-TextFieldRoot {
   display: flex;
-  box-sizing: border-box;
-  position: relative;
-  z-index: 0;
-}
+  align-items: stretch;
 
-.rt-TextFieldInput {
-  display: block;
-  width: 100%;
-  position: relative;
-  z-index: 1;
+  @supports selector(:has(*)) {
+    &:where(:has(.rt-TextFieldInput:focus)) {
+      outline: 2px solid var(--text-field-focus-color);
+      outline-offset: -1px;
+    }
+  }
+  @supports not selector(:has(*)) {
+    &:where(:focus-within) {
+      outline: 2px solid var(--text-field-focus-color);
+      outline-offset: -1px;
+    }
+  }
 
-  /* Clip text to the inner shadow */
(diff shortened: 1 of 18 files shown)


packages/radix-ui-themes/src/components/text-field.css
```diff
@@ -1,45 +1,104 @@
...
+  font-weight: var(--font-weight-regular);
+
+  @supports selector(:has(*)) {
+    &:where(:has(.rt-TextFieldInput:focus)) {
+      outline: 2px solid var(--text-field-focus-color);
+      outline-offset: -1px;
+    }
+  }
+  @supports not selector(:has(*)) {
+    &:where(:focus-within) {
```
benoitgrelard, line 15: Why not only do it via focus-within? It has better support than `:has` anyway. Or is that because if has wasn't supported and we'd use this, it would get focused when buttons in slots are focused?

benoitgrelard (commented):
Just want to flag these visual changes, you may be aware but just thought I'd point out, some very slight changes in where the text starts but also bigger changes in width for the ones at the bottom right.

https://github.com/radix-ui/themes/assets/1539897/50ba83e1-84ae-4eff-859e-49eec46922b2

vladmoroz:
> bigger changes in width for the ones at the bottom right

Fixed, I missed how those inputs were breaking out of the parent containers thinking that I actually had fixed them shrinking too much 😅 
https://github.com/radix-ui/themes/pull/291/commits/3e6208464341cd5bf7beb1d46cb2982efd01f759

benoitgrelard (approved):
There's still a slight difference when comparing to before, we used to have a 1px transparent border on the input which made the whole thing a little bigger overall. Not sure if that was a mistake before and we fixed it now by removing it or what.

vladmoroz:
> There's still a slight difference when comparing to before, we used to have a 1px transparent border on the input which made the whole thing a little bigger overall. Not sure if that was a mistake before and we fixed it now by removing it or what.

Yeah it's because I re-allocated that 1px space onto the root part. I think we shouldn't worry about it, it's visible only in edge case when you force the intrinsic size of the input (e.g. via `width: min-content` on the root or flex column with `align-items: start`; and in this case it's different between browsers anyway).

Also just committed an update to fix missing submenu icon color and [this regression](https://work-os.slack.com/archives/C066T3GNG4Q/p1707480993517169)
https://github.com/radix-ui/themes/pull/291/commits/9856253ccb4f4aca030678320fb45b629f471016

Could you check that too and let me know if you have any nits?

benoitgrelard (approved):

PR #298, opened 2024-02-14
Make width and height prop values out of scale

vladmoroz:
The `width` and `height` props won't map to space scale anymore. This is because in the vast majority of cases, width and height were not set to space scale, and with that, space scale as an IDE autocomplete suggestion felt odd/misleading.


packages/radix-ui-themes/changelog.md
```diff
@@ -24,6 +24,8 @@
     - `flexBasis`, `flexShrink`, `flexGrow`
     - `gridColumn`, `gridColumnStart`, `gridColumnEnd`
     - `gridRow`, `gridRowStart`, `gridRowEnd`
+  - **[Breaking]** The `width` and `height` props don't map to space scale anymore. This is because in the vast majority of cases, width and height were not set to space scale, and with that, space scale as an IDE autocomplete suggestion felt odd/misleading.
+    - **[Upgrade guide]** Find and replace your `width` and `height` prop usage with the corresponding [space scale](https://github.com/radix-ui/themes/blob/main/packages/radix-ui-themes/src/styles/tokens/space.css) steps, e.g. `width="9"` would need to be changed to `width="64px"`, and so on for other steps.
```
benoitgrelard, line 28: Should we suggest they could also refer to the CSS variables?

packages/radix-ui-themes/src/components/container.tsx
```diff
@@ -2,8 +2,14 @@ import * as React from 'react';
 import classNames from 'classnames';
 import { Slot } from '@radix-ui/react-slot';
 import { containerPropDefs } from './container.props.js';
-import { extractProps, getRoot } from '../helpers/index.js';
-import { deprecatedLayoutPropDefs, layoutPropDefs, marginPropDefs } from '../props/index.js';
+import { extractProps, getRoot, subtract } from '../helpers/index.js';
+import {
+  deprecatedLayoutPropDefs,
```
benoitgrelard, line 7: Just checking, are we going to remove the deprecated stuff now that we're making a major version?

packages/radix-ui-themes/src/components/container.tsx
```diff
@@ -15,36 +21,46 @@ interface ContainerProps
...
+  ({ width, minWidth, maxWidth, ...props }, forwardedRef) => {
+    const {
+      asChild,
+      children: childrenProp,
+      className,
+      ...containerProps
+    } = extractProps(
+      props,
+      containerPropDefs,
+      subtract(layoutPropDefs, widthPropDefs),
```
benoitgrelard, line 34: why do we need to do this?

packages/radix-ui-themes/src/styles/utilities/height.css
benoitgrelard, line 1: we need to remove those imports somewhere probably right?

vladmoroz: Imports? This file still has contents, I just removed the hardcoded space scale values CSS

packages/radix-ui-themes/changelog.md
vladmoroz: Good point, will add!

packages/radix-ui-themes/src/components/container.tsx
vladmoroz: Yep, just separately

packages/radix-ui-themes/src/components/container.tsx
vladmoroz: I guess we don't need to with width props having no default right now, but we'd need to if these prop defs were setting a default value. Should we remove and use this only as needed?

benoitgrelard (approved):

PR #299, opened 2024-02-15
Rework Text Field parts

vladmoroz:
**Breaking**
Remove the `Input` part to simplify how props are forwarded and rework internal HTML structure and styles.

**Upgrade guide**
  - All `TextField.Input` parts used **without** `TextField.Root` should be renamed to `TextField.Root`
  - All `TextField.Input` parts used **within** `TextField.Root` should be removed and their props should be put directly on the `TextField.Root` part
  - All `TextField.Slot` parts placed to the right of `TextField.Input` will need `side="right"` prop. However, no adjustment is needed when two slots were used within one `TextField`, e.g. one slot on the left and another one on the right. In that case, the slots will be automatically placed on different sides.
  - Make sure that your code works as expected if you were relying on any of the implementation quirks to override styles or behaviour.


packages/radix-ui-themes/src/components/text-field.tsx
```diff
@@ -2,43 +2,55 @@
...
+    | 'hidden'
+    | 'month'
+    | 'number'
+    | 'password'
+    | 'search'
+    | 'tel'
+    | 'text'
+    | 'time'
+    | 'url'
+    | 'week';
```
vladmoroz, line 28: This includes every input type where you can type characters, plus `hidden`

packages/radix-ui-themes/src/components/text-field.css
```diff
@@ -368,14 +379,15 @@
     }
     &:where(:placeholder-shown) {
       cursor: var(--cursor-disabled);
```
benoitgrelard, line 381: What's the logic behind only showing the disabled cursor if it's a placeholder and not if there's a value in the field?

packages/radix-ui-themes/src/components/text-field.css
```diff
@@ -120,10 +120,22 @@
   &:where(:empty) {
     display: none;
   }
-  &:where(:first-child) {
+
+  /* A slot that is not a right-side slot goes on the left */
+  &:where(:not([data-side='right'])) {
+    order: -1;
     margin-left: calc(var(--text-field-border-width) * -1);
+    margin-right: 0;
```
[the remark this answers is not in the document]
vladmoroz: We could drop this `margin-right: 0` but not `margin-left: 0` on the other one

packages/radix-ui-themes/src/components/text-field.css
vladmoroz: If there's a value in the field, you can still select and copy it, but when there's a placeholder, there's no interaction you can do.

benoitgrelard (approved):

PR #304, opened 2024-02-19
Single div cards

vladmoroz:
- Make card components single div
- Make radio and checkbox cards work with inset and allow them to blend border with parent background


packages/radix-ui-themes/src/components/base-card.css
```diff
@@ -0,0 +1,202 @@
...
+   * 1. "z-index: -1" so that the background goes below the children
+   * 2. "contain: paint" creates a new stacking context so that ::before doesn’t go below the card’s sibling elements
+   */
+  &::before {
+    z-index: -1;
+  }
+  & {
+    contain: paint;
+  }
+}
```
benoitgrelard, line 56: 🤯

packages/radix-ui-themes/src/components/card.css
```diff
@@ -115,148 +58,123 @@
...
+  --margin-bottom-override: calc(var(--margin-bottom) - var(--card-padding));
+  --margin-left-override: calc(var(--margin-left) - var(--card-padding));
+
+  margin-top: var(--margin-top-override);
+  margin-right: var(--margin-right-override);
+  margin-bottom: var(--margin-bottom-override);
+  margin-left: var(--margin-left-override);
+
+  /* Reset the overrides */
+  .rt-CardInner {
```
benoitgrelard, line 172: `rt-CardInner` doesn't exist anymore right as it's one element.
How do we ensure the margin overrides are reset for children?

vladmoroz: Oh right, good spot. Updated to `:where(&) > *` like we do in other instances

benoitgrelard (approved):

PR #305, opened 2024-02-19
Add Segmented Control

vladmoroz:
Screenshot 2024-02-19 at 14 05 29

### DIFF
--- a/apps/playground/app/sink/page.tsx
+++ b/apps/playground/app/sink/page.tsx
@@ -5915,7 +5915,7 @@ export default function Sink() {
                             </tr>
                           </thead>
                           <tbody>
-                            {selectTriggerPropDefs.radius.values.map((radius) => (
+                            {segmentedControlRootPropDefs.radius.values.map((radius) => (
                               <tr key={radius}>
                                 <RowHeaderCell>{radius}</RowHeaderCell>
                                 {segmentedControlRootPropDefs.size.values.map((size) => (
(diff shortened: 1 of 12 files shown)


apps/playground/app/sink/page.tsx
```diff
@@ -5858,6 +5861,183 @@ export default function Sink() {
...
+                          <thead>
+                            <tr>
+                              <ColumnHeaderCell />
+                              {segmentedControlRootPropDefs.size.values.map((size) => (
+                                <ColumnHeaderCell key={size}>size {size}</ColumnHeaderCell>
+                              ))}
+                            </tr>
+                          </thead>
+                          <tbody>
+                            {selectTriggerPropDefs.radius.values.map((radius) => (
```
benoitgrelard, line 5918:
```suggestion
                            {segmentedControlRootPropDefs.radius.values.map((radius) => (
```

benoitgrelard (commented):
I find the hover state a bit odd when there's some radius, it gets worse the bigger the radius. I wonder if the hover should follow the "shape" of the indicator (ie. have the same radius on all corners.

vladmoroz:
> I find the hover state a bit odd when there's some radius, it gets worse the bigger the radius. I wonder if the hover should follow the "shape" of the indicator (ie. have the same radius on all corners.

I wrestled a bit with this but found that inheriting the radius throughout feels worse for everything overall, and not 10x better even for full radius, which still looked odd:

Screenshot 2024-02-19 at 16 47 07

The separators is what makes the full radius feel slightly weird, but I don't think it worth it to special-case that radius and remove the separators. I'd keep this as is, as for all the other radii the hover that matches the separator feels better to me.

benoitgrelard (approved):

PR #307, opened 2024-02-19
Lint JS

vladmoroz:


benoitgrelard (approved):
Do we need to add some CI action for this?

vladmoroz:
@benoitgrelard don't think we have to, we lint pre-release anyway.

PR #310, opened 2024-02-20
Nail Data List alignment

vladmoroz:
- Update Data List so that the Value part is allowed to be a couple pixels higher than the label. This way small buttons and badges used in the Data List won't affect the layout and keep the visual gap between the items consistent.
- Allow text trunctation in out of the box

Comparison:
- [Before](https://themes-playground-lml63qlo8-modulz.vercel.app/test-data-list)
- [After](https://themes-playground-git-vlad-data-list-alignment-modulz.vercel.app/test-data-list)

image


packages/radix-ui-themes/src/components/data-list.css
```diff
@@ -19,6 +60,13 @@
 @breakpoints {
   .rt-DataListRoot {
     &:where(.rt-r-orientation-vertical) {
+      --data-list-value-margin-top: 0px;
+      --data-list-value-margin-bottom: 0px;
+      --data-list-first-item-label-margin-top: var(--leading-trim-start);
+      --data-list-last-item-label-margin-bottom: 0px;
+      --data-list-first-item-value-margin-top: 0px;
+      --data-list-last-item-value-margin-bottom: var(--leading-trim-end);
```
vladmoroz, line 68: We don't _have_ to have a billion CSS variables here, but given the complexity of those selectors across orientation x trim x first/last-child cases, this seems easier to understand and is better when it comes to the amount of the compiled CSS which `@breakpoints` yields

packages/radix-ui-themes/src/props/height.props.ts
```diff
@@ -1,7 +1,5 @@
 import { GetPropDefTypes, PropDef } from './prop-def.js';
 
-const heightValues = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'] as const;
```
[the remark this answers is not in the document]
vladmoroz: Yeah should have deleted it when we removed the scale values

vladmoroz:
OK, damn, started to respond how it works and realized the margin situation doesn't cooperate with align-items. Will need another good pass here.

packages/radix-ui-themes/src/components/data-list.css
```diff
@@ -1,9 +1,3 @@
-.rt-DataListItem {
```
benoitgrelard: DId you mean to delete this?

vladmoroz: Yeah, this stuff is set in the orientation block

benoitgrelard (approved):

PR #311, opened 2024-02-21
Simplify complex `asChild` scenarios and fix reconciliation issues

benoitgrelard:
## Description

This PR replaces the `getRoot` helper with a new helper that is simpler and doesn't have the caveats of creating a new component on each render. This should fix the issues observed by @lucasmotta.


packages/radix-ui-themes/src/components/tab-nav.tsx
```diff
@@ -46,27 +46,26 @@ interface TabNavLinkProps
...
         className={classNames('rt-reset', 'rt-BaseTabListTrigger', 'rt-TabNavLink', className)}
         onSelect={() => {}}
+        asChild={asChild}
       >
-        <span className="rt-BaseTabListTriggerInner rt-TabNavLinkInner">{children}</span>
-        <span className="rt-BaseTabListTriggerInnerHidden rt-TabNavLinkInnerHidden">
-          {children}
-        </span>
-      </TabNavLinkRoot>
+        {firstChildMightAdoptSubtree({ asChild, children }, (children) => (
```
benoitgrelard, line 60: I went with the render prop approach as I felt it was actually the simplest to understand the overall abstraction in one go whilst still retaining good ergonomics looking at the tree and colocation of the inner content.

The alternative I tested was to have 2 functions:
- one placed where `firstChildMightAdoptSubtree` currently is but taking a "node" as second argument…
- …in which you'd call a second like `getConsumerChildren({ asChild, children })` inline in your nested DOM

Would look something like this for this component:

```tsx
<NavigationMenu.Link
  {...linkProps}
  ref={forwardedRef}
  className={classNames('rt-reset', 'rt-BaseTabListTrigger', 'rt-TabNavLink', className)}
  onSelect={() => {}}
  asChild={asChild}
>
  {firstChildMightAdoptSubtree({ asChild, children },
    <>
      <span className="rt-BaseTabListTriggerInner rt-TabNavLinkInner">
        {getConsumerChildren({ asChild, children })}
      </span>
      <span className="rt-BaseTabListTriggerInnerHidden rt-TabNavLinkInnerHidden">
        {getConsumerChildren({ asChild, children })}
      </span>
    </>
  )}
</NavigationMenu.Link>
```

Note that this case needs 2 instances of the children so when I had that, I actually prefered abstracting the inner stuff in its own component so `getConsumerChildren` was only called once.

Overall, I think the render prop does a better job but let me know what you think!

packages/radix-ui-themes/src/helpers/first-child-might-adopt-subtree.ts
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * This is a helper function that is used when a component supports `asChild`
+ * using the `Slot` component but its implementation contains nested DOM elements.
+ *
+ * Using it ensures if a consumer uses the `asChild` prop, the elements are in
+ * correct order in the DOM, adopting the intended consumer `children`.
+ */
+export function firstChildMightAdoptSubtree(
+  options: { asChild: boolean | undefined; children: React.ReactNode },
+  content: React.ReactNode | ((children: React.ReactNode) => React.ReactNode)
```
benoitgrelard, line 12: Added the simple case where we don't need children as convenience (for cases like `Avatar` where the component actually doesn't take `children` (other than for `asChild` purposes).

packages/radix-ui-themes/src/helpers/first-child-might-adopt-subtree.ts
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+
+/**
+ * This is a helper function that is used when a component supports `asChild`
+ * using the `Slot` component but its implementation contains nested DOM elements.
+ *
+ * Using it ensures if a consumer uses the `asChild` prop, the elements are in
+ * correct order in the DOM, adopting the intended consumer `children`.
+ */
+export function firstChildMightAdoptSubtree(
```
vladmoroz, line 10: Do you have any other ideas for how to name this function? E.g. in a way that it would reflect what you get out of it

vladmoroz (approved):

PR #313, opened 2024-02-23
Add a wildcard export

vladmoroz:
Adding a wildcard export so that Next.js is able to resolve direct imports from the `dist` folder
https://work-os.slack.com/archives/C066T3GNG4Q/p1708683018800289

### DIFF
--- a/packages/radix-ui-themes/package.json
+++ b/packages/radix-ui-themes/package.json
@@ -37,26 +37,7 @@
         "default": "./dist/esm/props/index.js"
       }
     },
-    "./tokens.css": {
-      "import": "./tokens.css",
-      "require": "./tokens.css",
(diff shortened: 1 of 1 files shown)


packages/radix-ui-themes/package.json
```diff
@@ -37,26 +37,7 @@
...
-      "import": "./utilities.css",
-      "require": "./utilities.css",
-      "default": "./utilities.css"
-    },
-    "./styles.css": {
-      "import": "./styles.css",
-      "require": "./styles.css",
-      "default": "./styles.css"
-    }
+    "./*": "./*"
```
benoitgrelard, line 40: I assume you've double checked that the CSS files are resolved correctly?

benoitgrelard (approved):

PR #316, opened 2024-02-26
Add JSDoc for layout and Theme props

vladmoroz:
- Add JSDoc for layout and Theme props
- Enable esbuild’s minification (related as JSDoc comments added a lot of KB)


packages/eslint-config-custom/index.js
```diff
@@ -5,6 +5,7 @@ module.exports = {
     'jsx-a11y/alt-text': 'off',
     '@next/next/no-img-element': 'off',
     '@next/next/no-html-link-for-pages': 'off',
+    'react/jsx-curly-brace-presence': ['error', { props: 'never', children: 'ignore' }],
```
vladmoroz, line 8: Random, but was testing whether this can mitigate VSCode inserting `{}` for every responsive prop—seems to work well, would like to keep.

packages/radix-ui-themes/src/props/layout.props.ts
```diff
@@ -143,31 +322,10 @@ const layoutPropDefs = {
...
-} satisfies {
-  shrink: PropDef<(typeof flexShrinkValues)[number]>;
-  grow: PropDef<(typeof flexGrowValues)[number]>;
-};
-
-type LayoutProps = GetPropDefTypes<typeof layoutPropDefs & typeof deprecatedLayoutPropDefs>;
+// Use all of the imported prop defs to ensure that JSDoc works
+type LayoutProps = GetPropDefTypes<
+  typeof paddingPropDefs & typeof widthPropDefs & typeof heightPropDefs & typeof layoutPropDefs
+>;
```
vladmoroz, line 328: JSDoc won't work when we are just spreading padding, width, height, etc. prop defs into the layout pro defs; however it does work if we spread AND include the types here. This is the same reason `asChildProp` was changed to be an object—so that we can do `& typeof asChildPropDef`

benoitgrelard (commented):
Really nice work there @vladmoroz, that's going to improve the DX a lot! 👍

packages/radix-ui-themes/src/components/section.css
```diff
@@ -14,8 +14,12 @@
...
     }
     &:where(.rt-r-size-3) {
-      padding-top: 80px;
-      padding-bottom: 80px;
+      padding-top: var(--space-9);
+      padding-bottom: var(--space-9);
+    }
+    &:where(.rt-r-size-4) {
+      padding-top: calc(80px * var(--scaling));
+      padding-bottom: calc(80px * var(--scaling));
```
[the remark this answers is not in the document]
vladmoroz: Oh right, wanted to put this in a separate PR but missed it here. Yeah it kind of is breaking-ish? I'll add it to changelog. Felt like this size was always missing.

benoitgrelard (approved):

PR #322, opened 2024-02-27
Add width props to portalled components

vladmoroz:
`AlertDialog`, `Dialog`
- Add `width`, `minWidth`, `maxWidth` props to the Content part.
- Set `maxWidth="600px"` by default on the Content part.
  - **[Upgrade guide]** This slightly larger than the previous `580px` value. If you use dialogs that need a different width, override `maxWidth` with your own value.

***

`Popover`, `HoverCard`
- Add `width`, `minWidth`, `maxWidth`, `height`, `minHeight`, `maxHeight` props to the Content part.
- Set `maxWidth="480px"` by default on the Content part.
  - **[Upgrade guide]** If you use popovers and hover cards that need a wider width, override `maxWidth` with your own value.

***

`Tooltip`
- Add `width`, `minWidth`, `maxWidth` props.
- Set `maxWidth="200px"` by default on the tooltip content
  - **[Upgrade guide]** If you use tooltips that need to be wider, override `maxWidth` with your own value.

### DIFF
--- a/packages/radix-ui-themes/src/props/height.props.ts
+++ b/packages/radix-ui-themes/src/props/height.props.ts
@@ -1,6 +1,17 @@
 import { GetPropDefTypes, PropDef } from './prop-def.js';
 
 const heightPropDefs = {
+  /**
+   * Sets the CSS **height** property.
+   * Supports CSS strings and responsive objects.
+   *
+   * @example
+   * height="100px"
+   * height={{ md: '100vh', xl: '600px' }}
+   *
+   * @link
+   * https://developer.mozilla.org/en-US/docs/Web/CSS/height
+   */
   height: {
     type: 'string',
     className: 'rt-r-h',
@@ -8,6 +19,17 @@ const heightPropDefs = {
     default: undefined,
     responsive: true,
   },
+  /**
+   * Sets the CSS **min-height** property.
+   * Supports CSS strings and responsive objects.
+   *
+   * @example
+   * minHeight="100px"
+   * minHeight={{ md: '100vh', xl: '600px' }}
+   *
+   * @link
+   * https://developer.mozilla.org/en-US/docs/Web/CSS/min-height
+   */
   minHeight: {
     type: 'string',
     className: 'rt-r-min-h',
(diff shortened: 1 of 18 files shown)


packages/radix-ui-themes/src/props/prop-def.ts
```diff
@@ -68,10 +68,11 @@ type PropDef<T = any> = RegularPropDef<T> | ResponsivePropDef<T>;
...
 // prettier-ignore
 type GetPropDefType<Def> =
-	  Def extends BooleanPropDef ? (Def extends ResponsivePropDef ? Responsive<boolean> : boolean)
-  : Def extends StringPropDef ? (Def extends ResponsivePropDef ? Responsive<string> : string)
+    Def extends StringPropDef ? (Def extends ResponsivePropDef ? Responsive<string> : string)
+  : Def extends BooleanPropDef ? (Def extends ResponsivePropDef ? Responsive<boolean> : boolean)
   : Def extends ReactNodePropDef ? (Def extends ResponsivePropDef ? Responsive<React.ReactNode> : React.ReactNode)
-  : Def extends EnumOrStringPropDef<infer Type> ? (Def extends ResponsivePropDef<infer Type extends string> ? Responsive<Union<string, Type>> : Type)
+  : Def extends EnumOrStringPropDef<infer Type> ?
+    Def extends ResponsivePropDef<infer Type extends string> ? Responsive<Union<string, Type>> : Type
```
vladmoroz, line 75: We have something weird with this inference here. Was getting a responsive boolean type on `maxWidth` prop when it’s default was reassigned: 

Screenshot 2024-02-27 at 11 27 14

I couldn't find the root cause, but reordering the order in which the prop def type is inferred fixed that.

packages/radix-ui-themes/src/props/height.props.ts
```diff
@@ -1,20 +1,53 @@
 import { GetPropDefTypes, PropDef } from './prop-def.js';
 
 const heightPropDefs = {
+  /**
```
benoitgrelard, line 4: were those missed in the other PR? I can swear I'd seen the width/height stuff in that PR lol

packages/radix-ui-themes/src/props/prop-def.ts
benoitgrelard: I think I figured out what the issue is with this, let me know if you want to hop on a call.

benoitgrelard:
> I think I figured out what the issue is with this, let me know if you want to hop on a call.

I've just pushed the fix, but let me know if you'd prefer to catchup on it.

benoitgrelard (approved):

PR #335, opened 2024-03-04
Export colors and add functional steps

vladmoroz:
Review without whitespace


packages/radix-ui-themes/src/styles/tokens/base.css
vladmoroz, line 1: I went back and forth base vs core, landed on base because that’s how we call some other things in this project (base button, base dialog, etc). Imagining a potential future where base components or their styles are exported, this should make more sense altogether.

packages/radix-ui-themes/src/styles/tokens/color.css
```diff
@@ -282,332 +126,153 @@
...
+  --accent-3: var(--amber-3);
+  --accent-4: var(--amber-4);
+  --accent-5: var(--amber-5);
+  --accent-6: var(--amber-6);
+  --accent-7: var(--amber-7);
+  --accent-8: var(--amber-8);
+  --accent-9: var(--amber-9);
+  --accent-10: var(--amber-10);
+  --accent-11: var(--amber-11);
+  --accent-12: var(--amber-12);
```
vladmoroz, line 155: The steps still have to remapped here in `color.css`, otherwise components that use a color that’s not imported will fall back to the latest defined `--accent` variables, which felt weird.

vladmoroz: and tbh I believe there was another issue with that but I forgot what it was 😅

vladmoroz: The remaps are about 30 KB unminified ungzipped

packages/radix-ui-themes/src/styles/tokens/colors/yellow.css
```diff
@@ -0,0 +1,36 @@
...
+
+:root {
+  --yellow-contrast: #21201c; /* light mode sand 12 */
+}
+
+:root,
+.light,
+.light-theme {
+  --yellow-surface: #fefbe4cc;
+  --yellow-indicator: var(--yellow-10);
```
vladmoroz, line 14: Just yellow uses step 10 for the indicator, the rest use step 9

packages/radix-ui-themes/src/styles/tokens/color.css
benoitgrelard: hmm, that's a bit annoying, I wonder if we could do it differently with a fallback instead? So if not defined, it would fallback to some annoying color (would also help realising you need to import a missing color too).

packages/radix-ui-themes/src/styles/tokens/colors/lime.css
```diff
@@ -0,0 +1,36 @@
...
+  --lime-indicator: var(--lime-9);
+  --lime-track: var(--lime-9);
+  @supports (color: color(display-p3 1 1 1)) {
+    @media (color-gamut: p3) {
+      --lime-surface: color(display-p3 0.1098 0.1255 0.0784 / 0.5);
+    }
+  }
+  @supports (color: color-mix(in oklab, white, black)) {
+    --lime-track: color-mix(in oklab, var(--lime-8), var(--lime-9) 65%);
+  }
```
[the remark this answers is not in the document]
benoitgrelard: Which ones have been modified? The only one I can see a difference visually when comparing `Slider` in the sink is yellow.

packages/radix-ui-themes/src/styles/tokens/color.css
vladmoroz: OK I remembered the real reason actually, it's needed for realiasing colors

packages/radix-ui-themes/src/styles/tokens/colors/lime.css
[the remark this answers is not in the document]
vladmoroz: Compare dark mode

PR #337, opened 2024-03-05
Inherit text alignment for Text Field Input

vladmoroz:


packages/radix-ui-themes/src/components/reset.css
vladmoroz, line 1: Updated the tags that shouldn't inherit text-indent

benoitgrelard (approved):

PR #338, opened 2024-03-05
Fix theme panel `undefined` key errors in dev

vladmoroz:


packages/radix-ui-themes/src/components/theme-panel.tsx
```diff
@@ -135,7 +135,7 @@ const ThemePanelImpl = React.forwardRef<ThemePanelImplElement, ThemePanelImplPro
       function handleKeydown(event: KeyboardEvent) {
         const isModifierActive = event.altKey || event.ctrlKey || event.shiftKey || event.metaKey;
         const isKeyboardInputActive = document.activeElement?.closest(keyboardInputElement);
-        const isKeyT = event.key.toUpperCase() === 'T' && !isModifierActive;
+        const isKeyT = event.key?.toUpperCase() === 'T' && !isModifierActive;
```
vladmoroz, line 138: TS doesn't think `key` could be undefined, but I'm definitely seeing keydown events without `key` property when using keyboard early on page load

benoitgrelard: wow, weird, I didn't think this was possible… We never check this in primitives, don't feel like that should be needed…

benoitgrelard (approved):

packages/radix-ui-themes/src/components/theme-panel.tsx
vladmoroz: Was also surprised. I get this on the first keypress in sink
image

benoitgrelard: weird, I don't see this.

PR #340, opened 2024-03-05
Update CSS exports to work with postcss-import

vladmoroz:


benoitgrelard (approved):
Should we also update the clean task?

vladmoroz:
> Should we also update the clean task?

✅

PR #347, opened 2024-03-08
Support type=color on TextFieldRoot

penx:
## Description

When using TextFieldRoot with [getInputProps](https://conform.guide/api/react/getInputProps) from [@conform-to/react](https://conform.guide), you get a TypeScript error as the props returned from getInputProps can return type='color'.

I think it's reasonable to allow [type=color](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color) to be set on a Radix TextInput, so fixing this by adjusting types to allow for this.

## Testing steps

The following should not produce a TypeScript error:

```tsx
import { getInputProps, useForm } from "@conform-to/react";
import { TextFieldRoot } from "@radix-ui/themes";

function MyComponent() {
  const [form, fields] = useForm({});

  return (
    <TextFieldRoot
      {...getInputProps(fields.name, { type: "text" })}
    ></TextFieldRoot>
  );
}

```

### DIFF
--- a/packages/radix-ui-themes/src/components/text-field.tsx
+++ b/packages/radix-ui-themes/src/components/text-field.tsx
@@ -13,6 +13,7 @@ import { composeRefs } from '@radix-ui/react-compose-refs';
 type TextFieldRootElement = React.ElementRef<'input'>;
 type TextFieldRootOwnProps = GetPropDefTypes<typeof textFieldRootPropDefs> & {
   type?:
+    | 'color'
     | 'date'
     | 'datetime-local'
     | 'email'


penx:
Admittedly the result when you use type=color isn't great, but it works and I don't think that's a reason to not include it in types, given potential type compatibility issues

Screenshot 2024-03-08 at 12 40 53

penx:
Note that this had been working in 2.0 and in some of the 3.0 prereleases but broke recently, maybe around prerelease 10-15.

I also realise I can probably do this as a workaround:

```tsx
   <TextFieldRoot
      {...getInputProps(fields.name, { type: "text" })}
      type="text"
    ></TextFieldRoot>
```

vladmoroz:
@penx yep I know exactly when this change was made; the `type` on Text Field was narrowed to the ones that make sense for this component and have been designed for.

The workaround makes sense though, and I wonder if `{...getInputProps(fields.name, { type: "text" })}` already implies it will be `type="text"` and whether the `getInputProps` function should do a better job on narrowing down its return type.

penx:
I think you're right, and this can probably be closed as type=color probably doesn't make sense for this component as-is.

vladmoroz:
Cool, I'll close it for now. We had some reservation about whether narrowing the input type would break certain cases like this so we'll see if anything else comes up down the line.

PR #352, opened 2024-03-11
Fix Switch, Checkbox, and Radio with Skeleton

vladmoroz:
Fix height and radius of Switch, Checkbox, and Radio when used with a Skeleton


benoitgrelard (approved):
Nice, didn't realise this had broke. I guess it did when we changed the implementation to one part?

vladmoroz:
> Nice, didn't realise this had broke. I guess it did when we changed the implementation to one part?

I don't think it worked

benoitgrelard:
> > Nice, didn't realise this had broke. I guess it did when we changed the implementation to one part?
> 
> I don't think it worked

It definitely worked when I added `Skeleton` so something must  have changed in the meantime: https://themes-playground-5yw9ng9jv-modulz.vercel.app/sink

vladmoroz:
> It definitely worked when I added `Skeleton` so something must have changed in the meantime: https://themes-playground-5yw9ng9jv-modulz.vercel.app/sink

I think what changed was just that the radius was removed from the root parts, which makes sense to me.

Otherwise, in the old implementation `--line-height: 30px` over a Radio that Skeleton'd looks like an ellipse:

image

PR #356, opened 2024-03-12
Only dot-notation

benoitgrelard:
This PR removes the fully qualified named exports for multi-parts components, thus preferring dot-notation usage throughout.

cc @penx


penx (approved):
Perhaps component display names should be updated too. May be value in doing that later though, to isolate any potential (but I think unlikely) issues.

packages/radix-ui-themes/src/components/checkbox-card-group.tsx
```diff
@@ -117,5 +117,5 @@ const CheckboxCardGroupItem = React.forwardRef<
 });
 CheckboxCardGroupItem.displayName = 'CheckboxCardGroupItem';
 
-export { CheckboxCardGroupRoot, CheckboxCardGroupItem };
-export type { CheckboxCardGroupRootProps, CheckboxCardGroupItemProps };
+export { CheckboxCardGroupRoot as Root, CheckboxCardGroupItem as Item };
```
vladmoroz, line 120: Turbo nit, but does Prettier coerce us into doing everything on a single line here? I'd prefer multi-line formatting here for consistency and readability

vladmoroz:
> What do you think @vladmoroz? That might make sense given what we have discussed earlier in terms of the parallel between single/multi parts components import/types (and now display names)?

I don't mind, don't have a preference here

packages/radix-ui-themes/src/components/checkbox-card-group.tsx
benoitgrelard: prettier ignore?

vladmoroz: Damn yeah it doesn't, never mind. Not willing to prettier ignore it all. Just thought it's one of those cases when you can do a line break and it'd respect it.

benoitgrelard: This works

CleanShot 2024-03-12 at 14 52 29@2x

vladmoroz: Naah not worth it

PR #363, opened 2024-03-13
Normalize prop def shapes

vladmoroz:


packages/radix-ui-themes/src/components/avatar.props.ts
```diff
@@ -1,23 +1,25 @@
...
+  highContrastPropDef,
+  radiusPropDef,
+} from '../props/index.js';
 import type { PropDef } from '../props/index.js';
 
 const sizes = ['1', '2', '3', '4', '5', '6', '7', '8', '9'] as const;
 const variants = ['solid', 'soft'] as const;
 
 const avatarPropDefs = {
-  ...asChildProp,
```
vladmoroz, line 8: I went 180° on my proposal to rename all `componentPropDefs` to `componentProps` and instead renamed all prop defs that didn't have the suffix because that added some conflicts in JSX when we renamed props when spreading (e.g. `asChild: asChildProp`)

packages/radix-ui-themes/src/components/progress.props.ts
```diff
@@ -1,23 +1,19 @@
-import { colorProp, highContrastProp, radiusProp, widthPropDefs } from '../props/index.js';
+import { colorPropDef, highContrastPropDef, radiusPropDef, widthPropDefs } from '../props/index.js';
 import type { PropDef } from '../props/index.js';
 
 const sizes = ['1', '2', '3'] as const;
 const variants = ['classic', 'surface', 'soft'] as const;
 
 const progressPropDefs = {
-  ...widthPropDefs,
```
vladmoroz, line 8: Removed width prop from Progress

packages/radix-ui-themes/src/components/avatar.props.ts
benoitgrelard: haha, there's always something we don't account for 😀

benoitgrelard (approved):

PR #364, opened 2024-03-13
Remove explore components page

vladmoroz:


benoitgrelard (approved):
is the idea to only maintain this in the website repo?

vladmoroz:
> is the idea to only maintain this in the website repo?

Yeah I think it’s not too sustainable to maintain it in two places

PR #388, opened 2024-03-18
Minimize usage of default utility classes values

vladmoroz:
- Resolves an edge case when utility classes CSS is imported on top of custom styles and take over the custom styles unexpectedly when building on top of certain components
- Add missing box-sizing to components that have padding props on them

### DIFF
--- a/packages/radix-ui-themes/src/components/base-tab-list.props.ts
+++ b/packages/radix-ui-themes/src/components/base-tab-list.props.ts
@@ -18,7 +18,7 @@ const baseTabListPropDefs = {
     type: 'enum',
     className: 'rt-r-jc',
     values: justifyValues,
-    default: 'start',
+    default: undefined,
     responsive: true,
   },
   ...colorPropDef,
--- a/packages/radix-ui-themes/src/components/container.tsx
+++ b/packages/radix-ui-themes/src/components/container.tsx
@@ -15,7 +15,7 @@ interface ContainerProps
     LayoutProps,
     ContainerOwnProps {}
 const Container = React.forwardRef<ContainerElement, ContainerProps>(
-  ({ width, minWidth, maxWidth, ...props }, forwardedRef) => {
+  ({ width, minWidth, maxWidth, height, minHeight, maxHeight, ...props }, forwardedRef) => {
     const { asChild, children, className, ...containerProps } = extractProps(
       props,
       containerPropDefs,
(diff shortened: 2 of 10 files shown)


packages/radix-ui-themes/src/components/container.tsx
```diff
@@ -24,7 +24,7 @@ const Container = React.forwardRef<ContainerElement, ContainerProps>(
     );
 
     const { className: innerClassName, style: innerStyle } = extractProps(
-      { width, minWidth, maxWidth },
+      { width, minWidth, maxWidth, height, minHeight, maxHeight },
```
vladmoroz, line 27: Felt weird that width style props go on the inner element, but not height. If container root gets a height, the inner element won't stretch vertically to fill it.

packages/radix-ui-themes/src/components/flex.props.ts
```diff
@@ -33,7 +33,7 @@ const flexPropDefs = {
     type: 'enum',
     className: 'rt-r-display',
     values: displayValues,
-    default: 'flex',
+    default: undefined,
```
vladmoroz, line 36: Already in the styles

packages/radix-ui-themes/src/components/flex.props.ts
```diff
@@ -88,7 +88,7 @@ const flexPropDefs = {
     className: 'rt-r-jc',
     values: justifyValues,
     parseValue: parseJustifyValue,
-    default: 'start',
+    default: undefined,
```
vladmoroz, line 91: Already in the styles

packages/radix-ui-themes/src/components/base-tab-list.props.ts
```diff
@@ -18,7 +18,7 @@ const baseTabListPropDefs = {
     type: 'enum',
     className: 'rt-r-jc',
     values: justifyValues,
-    default: 'start',
+    default: undefined,
```
benoitgrelard, line 21: Can we just omit those instead of explicitly setting to `undefined`?

vladmoroz: I believe so, do we want to do that for all prop defs?

benoitgrelard: If it works yeah I think so, unless there's a particular reason why we'd want that.

vladmoroz: I'll do it separately then in another PR on all prop defs to keep the history clean?

PR #389, opened 2024-03-18
Enforce typography defaults instead of inheriting

vladmoroz:


packages/radix-ui-themes/src/components/avatar.css
```diff
@@ -17,6 +17,9 @@
 }
 
 .rt-AvatarFallback {
+  font-family: var(--default-font-family);
```
benoitgrelard, line 20: What issues does it solve?

vladmoroz: I've ran into a case when I put a floating button into a `pre` and that overrode its font

Thinking through this it felt unexpected that the font was inherited; imagine a situation where serif font blog post has a card in the middle to link out to another piece of content, potentially with a button inside. I think it would be expected that these self-contained components retain the default font

benoitgrelard (approved):

Issue #382, opened 2024-03-17
[Feature Request] Allow programmatically controlling the `ThemePanel` component


snelsi:
`ThemePanel` component is great, but I think it can be improved by allowing the user to control more of its behavior:

1) Allow passing `open` and `onOpenChange` to `ThemePanel` to control it's state
2) Allow customizing hotkeys for "show/hide" and "switch dark mode" actions. Also, allow passing `null` to disable a hotkey.

Here are a couple of great custom hotkey hooks for inspiration:
https://github.com/JohannesKlauss/react-hotkeys-hook
https://github.com/arthurtyukayev/use-keyboard-shortcut
https://github.com/greena13/react-hotkeys

vladmoroz:
Why are you interested in that? What are you trying to achieve in your app?

snelsi:
I want to add a custom button in the header/footer that toggles a `ThemePanel` on click. 
And I want to remap "T" and "D" hotkeys to something else. 

Also, when `ThemePanel` is closed It really misses some sort of floating button that reopens it.

vladmoroz:
I am more interested about this in terms of what this achieves for your users? Why do you need Theme Panel in production? It was never meant as a component that the end user would see; it's just a tool for us to demo the theme options better and for the developer to explore their app style.

snelsi:
I'm building a catalogue of premade components/sections. 
Very similar to [Radix Playground](https://www.radix-ui.com/themes/playground). 
I'm using it to demo the theme options better as well.

vladmoroz:
Feel free to PR `open` and `onOpenChange` props and a prop to disable hotkeys.

We won't want to add a way to reassign hotkeys because using Theme Panel in a real app is not a case that we see as legitimate/realistic; nor it makes sense to add a dependency for that.

snelsi:
Okay, I'll take a look today

snelsi:
@vladmoroz 
Here's a PR with discussed changes:
https://github.com/radix-ui/themes/pull/393

PR #393, opened 2024-03-20
Feature/382 refactor theme panel component

snelsi:
### Main changes:

- Allow passing optional `open` and `onOpenChange` props to the `ThemePanel` component using the `useControllableState` hook
- Simplify setup by merging `ThemePanel` and `ThemePanelImpl` components. Remove redundant interfaces
- Move common logic related to hotkeys into a custom `useHotKey` hook _(no external dependencies)_
- Allow overriding default open/appearance hotkeys using optional `openHotkey` and `toogleAppearanceHotkey` props
- Allow passing `null` or `false` to disable a hotkey
- Hide the top-right open hotkey button when `openHotkey` is null

### Things to improve in the future:

- Allow passing combination of keys like `Control + F` or `Shift + H`
- `useHotKey` implementation is very barebones. Consider using a library like [react-hotkeys-hook](https://github.com/JohannesKlauss/react-hotkeys-hook)

## Testing steps

Use cases:

1) Controlled behaviour:
```tsx
const [open, setOpen] = useState(false);

return <ThemePanel open={open} onOpenChange={setOpen} />
```

2) Custom hotkeys:
```tsx
return <ThemePanel openHotkey="P" toogleAppearanceHotkey="/" />
```

3) Disabling hotkeys:
```tsx
return <ThemePanel openHotkey={null} toogleAppearanceHotkey={false} />
```

## Relates issues / PRs

Refs #382

### DIFF
--- a/packages/radix-ui-themes/src/components/theme-panel.tsx
+++ b/packages/radix-ui-themes/src/components/theme-panel.tsx
@@ -2,6 +2,7 @@
 
 import * as React from 'react';
 import { useCallbackRef } from '@radix-ui/react-use-callback-ref';
+import { useControllableState } from '@radix-ui/react-use-controllable-state';
 import {
   AccessibleIcon,
   Box,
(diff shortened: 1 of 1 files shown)


packages/radix-ui-themes/src/components/theme-panel.tsx
```diff
@@ -23,605 +24,628 @@ import { themePropDefs } from '../props/index.js';
...
+  /** A hotkey to quickly show/hide the panel.
+   * Set to `null` or `false` to disable the hotkey.
+   * @default "T"
+   */
+  openHotkey?: string | null | false;
+  /** A hotkey to quickly toggle the appearance.
+   * Set to `null` or `false` to disable the hotkey.
+   * @default "D"
+   * */
+  toogleAppearanceHotkey?: string | null | false;
```
vladmoroz, line 84: We wouldn't want to expose a way to customize the shortcut itself – the theme panel should always work the same in different places. Having different shortcuts on different websites is going to feel unpredictable.

Just a boolean prop to disable the shortcuts should be fine, e.g. `disableShortcuts`

packages/radix-ui-themes/src/components/theme-panel.tsx
```diff
@@ -23,605 +24,628 @@ import { themePropDefs } from '../props/index.js';
...
+
+/** Listen to keydown events and fire a callback when a hotkey is pressed */
+const useHotKey = (
+  key: string | null | false,
+  callback: () => void,
+  options?: {
+    /** Determines whether the hotkey listener is active or not */
+    enabled?: boolean;
+  }
+) => {
```
vladmoroz, line 45: Let's put the hook and the input elements selector back into the component body; it's easier to read code when things are declared close to where they are used

snelsi: It doesn't make sense to inline this effect into the component's body because you use it in two different places. You'll end up with two chunks of the same code.

Plus, it's located in the same file, just above where it's used, and the name is self-explanatory in this case. That should be good enough.

I can move the `keyboardInputElement` declaration into the useHotKey hook's body if you want.

vladmoroz: > It doesn't make sense to inline this effect into the component's body because you use it in two different places. You'll end up with two chunks of the same code.

Oh I see, you are right, I missed that

packages/radix-ui-themes/src/components/theme-panel.tsx
snelsi: I respectfully disagree with your perspective. Default behavior can indeed remain consistent, but offering customization options allows users to adapt the component to fit seamlessly into their workflows. If a user chooses to change a defined hotkey, it should be within their responsibility, and providing such flexibility enhances the usability and adaptability of the library.

snelsi: By the way, just noticed that Radix Primitives has an example of [Custom hotkey](https://www.radix-ui.com/primitives/docs/components/toast#custom-hotkey) prop for Toast component 🙂

image

chaance:
Hi! Just wanted to weigh in here since we've had a little maintainer shuffle since this PR was opened. 

Any chance we could break this into two separate, stacked PRs? I'd like to consider each new feature independently since there's a little unresolved feedback. Specifically it'd be nice to see it broken into three:
- Making `ThemePanel` use controllable state
- New `useHotKey` hook
- Allow for customizing/disabling hotkeys (and hiding the hotkey button accordingly)

I know it's been a while, so totally understand if you've moved on. Happy to take this as a feature request and re-implement if we decide to move forward.

packages/radix-ui-themes/src/components/theme-panel.tsx
chaance: FWIW Radix Primitives is designed to be more flexible, where Themes is a more opinionated starter. I could be convinced on this, however. I agree it can be annoying to deal with potential conflicts.

PR #394, opened 2024-03-20
Relax polymorphic types

vladmoroz:
Relax polymorphic types to make extending components easier
https://work-os.slack.com/archives/C066T3GNG4Q/p1710869863537029


packages/radix-ui-themes/src/components/heading.tsx
```diff
@@ -5,19 +5,16 @@ import { headingPropDefs } from './heading.props.js';
 import { extractProps } from '../helpers/index.js';
 import { marginPropDefs } from '../props/index.js';
 
-import type { ComponentPropsWithout, NiceIntersection, RemovedProps } from '../helpers/index.js';
```
benoitgrelard, line 8: I think we can remove that type from the codebase now then?

benoitgrelard (approved):

PR #400, opened 2024-03-22
Changelog tweaks

benoitgrelard:
A few tweaks I made before/during adding the release to the website

### DIFF
--- a/packages/radix-ui-themes/changelog.md
+++ b/packages/radix-ui-themes/changelog.md
@@ -7,7 +7,8 @@
     - Improve ESM compatibility
     - Improve tree-shaking of individual component parts
     - **[Breaking]** Drop named exports for multi-part components
-      - **[Upgrade guide]** use dot notation instead, ie. `Dialog.Root`
+      - **[Upgrade guide]** Use dot notation for all multi-part components instead, ie. `Dialog.Root`
+      - Note: Our new approach allows dot notation to work reliably in server components too
     - **[Breaking]** Remove component prop definitions and internal helpers from the root `@radix-ui/themes` import entry point and export them from `@radix-ui/themes/props` and `@radix-ui/themes/helpers` to make it possible to build your own component library on top of Radix Themes using the same techniques.
       - Note: you might need to use `"moduleResolution": "NodeNext"` with your compiler to access these paths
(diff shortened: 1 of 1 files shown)


packages/radix-ui-themes/changelog.md
```diff
@@ -6,7 +6,8 @@
   - Package structure
     - Improve ESM compatibility
     - Improve tree-shaking of individual component parts
-    - Drop named exports for multi-part components (use dot notation instead, ie. `Dialog.Root`)
+    - **[Breaking]** Drop named exports for multi-part components
+      - **[Upgrade guide]** use dot notation instead, ie. `Dialog.Root`
```
vladmoroz, line 10:
```suggestion
      - **[Upgrade guide]** Use dot notation for all multi-part components instead, ie. `Dialog.Root`
```

We should also add that we fixed it not working with RSC

packages/radix-ui-themes/changelog.md
```diff
@@ -159,10 +161,10 @@
   - Add `color` and `highContrast` props to `TabsList`
   - Add margin props `TabsList` and `TabsContent`
   - Renamed the letter/word spacing CSS variables in `.radix-themes` so that it supports both `Tabs` and `TabNav` components. Note that this is a backwards compatible change, as the old variables are still supported.
-    - `--tabs-trigger-active-letter-spacing` -> `--tab-active-letter-spacing`
-    - `--tabs-trigger-active-word-spacing` -> `--tab-active-word-spacing`
-    - `--tabs-trigger-inactive-letter-spacing` -> `--tab-inactive-letter-spacing`
-    - `--tabs-trigger-inactive-word-spacing` -> `--tab-inactive-word-spacing`
+    - `--tabs-trigger-active-letter-spacing` → `--tab-active-letter-spacing`
```
vladmoroz, line 164: Above it says:
> Note that this is a backwards compatible change, as the old variables are still supported.

Not true anymore I believe, we got rid of the backwards compatibility?

benoitgrelard: Looks like it still exists
CleanShot 2024-03-22 at 14 38 30@2x

vladmoroz (approved):

PR #406, opened 2024-03-23
Work around bugged style linting

vladmoroz:


packages/radix-ui-themes/src/components/text.css
```diff
@@ -15,7 +15,7 @@
     color: var(--accent-12);
   }
 
-  /* stylelint-disable-next-line selector-max-type */
```
vladmoroz, line 18: Stylelint doesn't recognise this directive, probably because of how the styles are reshuffled into the media query

I think ultimately we should disable this rule and relying on the PR reviews to catch undesired type selectors, as it has been more annoying than helpful

PR #416, opened 2024-03-26
Fix Radio and Checkbox cursors in Safari

### LINKED ISSUE #413: RadioGroup disabled cursor style doesn't work in Safari
Before v3 the disabled cursor was applied directly to the radio group button but seems to have changed to use the ::before selector

https://github.com/radix-ui/themes/blob/481375cfbcf5d966ff6861936b716d9a45df4f4c/packages/radix-ui-themes/src/components/base-radio.css#L42

Making the following change brings back the previous behaviour

```diff
- &:where(:disabled, [data-disabled])::before { 
+ &:where(:disabled, [data-disabled]) { 
```

From the docs website
image

vladmoroz:
Resolves #413

### DIFF
--- a/packages/radix-ui-themes/src/components/base-checkbox.css
+++ b/packages/radix-ui-themes/src/components/base-checkbox.css
@@ -28,6 +28,11 @@
   position: absolute;
   width: var(--checkbox-indicator-size);
   height: var(--checkbox-indicator-size);
+
+  /* Required for subpixel alignment in Safari */
+  transform: translate(-50%, -50%);
+  top: 50%;
+  left: 50%;
 }
 
 /***************************************************************************************************
--- a/packages/radix-ui-themes/src/components/base-radio.css
+++ b/packages/radix-ui-themes/src/components/base-radio.css
@@ -15,6 +15,10 @@
   border-radius: var(--skeleton-radius);
   --skeleton-radius-override: 100%;
 
+  &:where(:disabled, [data-disabled]) {
+    cursor: var(--cursor-disabled);
+  }
+
   &::before {
     content: '';
     display: block;
@@ -37,10 +41,6 @@
     &::after {
       content: '';
     }
-  }
-
-  &:where(:disabled, [data-disabled])::before {
-    cursor: var(--cursor-disabled);
   }
 
   &:where(:focus-visible)::before {
--- a/packages/radix-ui-themes/src/components/checkbox-group.css
+++ b/packages/radix-ui-themes/src/components/checkbox-group.css
@@ -13,13 +13,18 @@
 }
 
 .rt-CheckboxGroupItemCheckbox {
-  &:where(:focus-visible)::before {
-    outline: 2px solid var(--focus-8);
-    outline-offset: 2px;
+  &:where(:focus-visible) {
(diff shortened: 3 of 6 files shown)


packages/radix-ui-themes/src/components/base-checkbox.css
```diff
@@ -28,6 +28,11 @@
   position: absolute;
   width: var(--checkbox-indicator-size);
   height: var(--checkbox-indicator-size);
+
+  /* Required for subpixel alignment in Safari */
+  transform: translate(-50%, -50%);
+  top: 50%;
+  left: 50%;
```
vladmoroz, line 35: Current situation:

Screenshot 2024-03-26 at 12 59 38

packages/radix-ui-themes/src/components/base-radio.css
```diff
@@ -15,6 +15,10 @@
   border-radius: var(--skeleton-radius);
   --skeleton-radius-override: 100%;
 
+  &:where(:disabled, [data-disabled]) {
+    cursor: var(--cursor-disabled);
+  }
```
vladmoroz, line 20: - The `::before` pseudo-element is currently used to render the checkbox and radio chrome, and the root targets the clickable area, which may be a bit larger that the visible element in certain layouts
- `cursor` styles is Safari don't apply to pseudo-elements (seems to be a relatively well-known issue)
- Ideally, we’d work around that using a separate `span` element for the chrome and put the `cursor` style there too
- However, we can’t do the above with the `Radio` component which renders a regular `input`, so I decided to put the cursor style back onto the entire clickable area given that it isn't much larger in reasonable layout situations

packages/radix-ui-themes/src/components/checkbox-group.css
```diff
@@ -13,13 +13,18 @@
...
 .rt-CheckboxGroupItemCheckbox {
-  &:where(:focus-visible)::before {
-    outline: 2px solid var(--focus-8);
-    outline-offset: 2px;
+  &:where(:focus-visible) {
+    &::before {
+      outline: 2px solid var(--focus-8);
+      outline-offset: 2px;
+    }
   }
```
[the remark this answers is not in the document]
vladmoroz: I prefer the style when the pseudos are separated the way a component’s internal parts would be; feels clearer to me that it's targeting a subelement. Not a big deal, let me know what your call would be here.

packages/radix-ui-themes/src/components/base-checkbox.css
benoitgrelard: hmm weird… we didn't add a new size in 3.0 right?

packages/radix-ui-themes/src/components/checkbox-group.css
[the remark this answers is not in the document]
benoitgrelard: ah gotcha.

benoitgrelard (approved):

packages/radix-ui-themes/src/components/base-checkbox.css
vladmoroz: Nope

PR #423, opened 2024-04-01
Configure Autoprefixer browser support

vladmoroz:
Configured Autoprefixer to support browser versions from the last 2 years

This saves 17 KB of the built CSS and cleans up the web inspector experience by removing the following unnecessary transformations:
- `::-moz-placeholder`
- `::-moz-selection`
- `:-moz-any-link`
- `:-moz-read-only`
- `-moz-user-select`
- `-moz-column-gap`
- `-o-object-fit`

Examples:

### DIFF
--- a/packages/radix-ui-themes/package.json
+++ b/packages/radix-ui-themes/package.json
@@ -137,7 +137,7 @@
   "devDependencies": {
     "@types/react": "^18.2.53",
     "@types/react-dom": "^18.2.18",
-    "autoprefixer": "^10.4.16",
+    "autoprefixer": "^10.4.19",
     "esbuild": "^0.20.0",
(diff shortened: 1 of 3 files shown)


packages/radix-ui-themes/package.json
```diff
@@ -137,7 +137,7 @@
   "devDependencies": {
     "@types/react": "^18.2.53",
     "@types/react-dom": "^18.2.18",
-    "autoprefixer": "^10.4.16",
+    "autoprefixer": "^10.4.19",
```
vladmoroz, line 140: Changes since 10.4.16:

image

benoitgrelard (approved):

PR #424, opened 2024-04-01
Fix auto high-contrast link color regression

vladmoroz:
Fixes a regression from https://github.com/radix-ui/themes/pull/387 when links wouldn't become high-contrast automatically when used in colored text:

image

### DIFF
--- a/packages/radix-ui-themes/src/components/index.css
+++ b/packages/radix-ui-themes/src/components/index.css
@@ -1,8 +1,12 @@
 @import './reset.css';
-/* Import Skeleton before other components so that its default border-radius doesn’t have a higher specificity */
-@import './skeleton.css';
 @import './animations.css';
 @import './layout.css';
+
+/* Import Skeleton before other components so that its default border-radius doesn’t have a higher specificity */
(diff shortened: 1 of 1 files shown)


packages/radix-ui-themes/src/components/index.css
```diff
@@ -1,9 +1,13 @@
...
-/* Import Skeleton before other components so that its default border-radius doesn’t have a higher specificity */
-@import './skeleton.css';
 @import './animations.css';
 @import './layout.css';
 
+/* Import Skeleton before other components so that its default border-radius doesn’t have a higher specificity */
+@import './skeleton.css';
+
+/* Import Text before other components as it’s commonly extended */
+@import './text.css';
```
benoitgrelard, line 9: Not sure how I feel about this as a solution, what is the actual root cause?

vladmoroz: Not sure how it had worked before to be honest, this seems like an obviously problematic area for components that extend text and target the same selectors like `[data-accent-color]`

benoitgrelard (approved):

PR #469, opened 2024-04-22
Link: Fix overzealous automatic high-contrast colour

vladmoroz:
- Fixes a regression when links would automatically become high-contrast when that wasn't desired, e.g [WorkOS dashboard sign-in](https://dashboard.workos.com/signin):

  image

- Fixes a regression when nested high-contrast gray text would not use the correct selection background color:

  image

- Simplify some of the color prop mechanics

### DIFF
--- a/packages/radix-ui-themes/src/components/callout.tsx
+++ b/packages/radix-ui-themes/src/components/callout.tsx
@@ -55,16 +55,8 @@ type CalloutIconElement = React.ElementRef<'div'>;
 interface CalloutIconProps extends ComponentPropsWithout<'div', RemovedProps> {}
 const CalloutIcon = React.forwardRef<CalloutIconElement, CalloutIconProps>(
   ({ className, ...props }, forwardedRef) => {
-    const { color, size, highContrast } = React.useContext(CalloutContext);
     return (
-      <Text
-        asChild
-        color={color}
-        size={mapResponsiveProp(size, mapCalloutSizeToTextSize)}
-        highContrast={highContrast}
-      >
-        <div {...props} className={classNames('rt-CalloutIcon', className)} ref={forwardedRef} />
-      </Text>
+      <div {...props} className={classNames('rt-CalloutIcon', className)} ref={forwardedRef} />
     );
   }
 );
@@ -74,13 +66,11 @@ type CalloutTextElement = React.ElementRef<'p'>;
 type CalloutTextProps = ComponentPropsAs<typeof Text, 'p'>;
 const CalloutText = React.forwardRef<CalloutTextElement, CalloutTextProps>(
   ({ className, ...props }, forwardedRef) => {
-    const { color, size, highContrast } = React.useContext(CalloutContext);
+    const { size } = React.useContext(CalloutContext);
     return (
       <Text
         as="p"
         size={mapResponsiveProp(size, mapCalloutSizeToTextSize)}
-        color={color}
-        highContrast={highContrast}
         {...props}
         asChild={false}
         ref={forwardedRef}
--- a/packages/radix-ui-themes/src/components/callout.tsx
+++ b/packages/radix-ui-themes/src/components/callout.tsx
@@ -13,20 +13,17 @@ import { GetPropDefTypes, MarginProps } from '../props/index.js';
 
 type CalloutRootOwnProps = GetPropDefTypes<typeof calloutRootPropDefs>;
 
-type CalloutContextValue = CalloutRootOwnProps;
+type CalloutContextValue = { size?: CalloutRootOwnProps['size'] };
 const CalloutContext = React.createContext<CalloutContextValue>({});
 
 type CalloutRootElement = React.ElementRef<'div'>;
 interface CalloutRootProps
   extends ComponentPropsWithout<'div', RemovedProps>,
     MarginProps,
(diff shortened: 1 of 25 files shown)


packages/radix-ui-themes/src/components/callout.tsx
```diff
@@ -55,16 +55,8 @@ type CalloutIconElement = React.ElementRef<'div'>;
...
-    const { color, size, highContrast } = React.useContext(CalloutContext);
     return (
-      <Text
-        asChild
-        color={color}
-        size={mapResponsiveProp(size, mapCalloutSizeToTextSize)}
-        highContrast={highContrast}
-      >
-        <div {...props} className={classNames('rt-CalloutIcon', className)} ref={forwardedRef} />
-      </Text>
```
vladmoroz, line 67: Text seemed unnecessary here; it was really needed just for the line height values which make more sense in the CSS

packages/radix-ui-themes/src/components/callout.tsx
```diff
@@ -74,13 +66,11 @@ type CalloutTextElement = React.ElementRef<'p'>;
...
 const CalloutText = React.forwardRef<CalloutTextElement, CalloutTextProps>(
   ({ className, ...props }, forwardedRef) => {
-    const { color, size, highContrast } = React.useContext(CalloutContext);
+    const { size } = React.useContext(CalloutContext);
     return (
       <Text
         as="p"
         size={mapResponsiveProp(size, mapCalloutSizeToTextSize)}
-        color={color}
-        highContrast={highContrast}
```
vladmoroz, line 83: Was unnecessary

packages/radix-ui-themes/src/components/link.css
```diff
@@ -17,7 +17,7 @@
     outline-offset: 2px;
   }
 
-  :where([data-accent-color]:not(.radix-themes, .rt-high-contrast)) & {
+  :where([data-accent-color]:not(.radix-themes, .rt-high-contrast)) &:where([data-accent-color='']) {
```
vladmoroz, line 20: This is the actual Link fix

packages/radix-ui-themes/src/styles/tokens/color.css
```diff
@@ -72,7 +72,7 @@
 /* * * * * * * * * * * * * * * * * * * */
 
 .radix-themes,
-[data-accent-color]:where(:not([data-accent-color='gray'])) {
+[data-accent-color]:where(:not([data-accent-color=''], [data-accent-color='gray'])) {
```
vladmoroz, line 75: This fixes the selection color

packages/radix-ui-themes/src/components/text.tsx
```diff
@@ -28,7 +28,7 @@ const Text = React.forwardRef<TextElement, TextProps>((props, forwardedRef) => {
   } = extractProps(props, textPropDefs, marginPropDefs);
   return (
     <Slot
-      data-accent-color={color || undefined}
```
vladmoroz, line 31: This was a bit of a weird workaround for when callout passed an empty string here, I think the component shouldn't hijack `color=''` into `color={undefined}`

packages/radix-ui-themes/src/components/callout.tsx
benoitgrelard: How is the size handled now?

packages/radix-ui-themes/src/components/callout.tsx
benoitgrelard: - How is it unnecessary then? Where is color/high-contrast set then?
- If unnecessary, should we update the context definition?

packages/radix-ui-themes/src/components/link.css
benoitgrelard: When is it `''`?

vladmoroz: `data-accent-color` is an empty string whenever no color is passed and it uses the current accent (via Text)

What we want is auto high contrast when you are not setting an explicit `color` prop, which we broke at some point

packages/radix-ui-themes/src/components/callout.tsx
vladmoroz: See the CSS changes in callout.css, it's just one variable

packages/radix-ui-themes/src/components/callout.tsx
vladmoroz: The `data-accent-color` attribute and text color CSS is set on Callout Root anyway. See the data-accent-color repetition here (this is current):

image

vladmoroz: Updated the context 950bea0b9266c0fb237f4a4d3226ba0037d0ebf0

packages/radix-ui-themes/src/components/link.css
benoitgrelard: Yeah saw that in the other PR, thanks!

packages/radix-ui-themes/src/components/callout.tsx
benoitgrelard: Yeah good point, not sure why I did that in the first place, seems odd.

benoitgrelard (approved):

PR #470, opened 2024-04-22
Rename color prop defs for clarity

vladmoroz:
Spinning off from https://github.com/radix-ui/themes/pull/469, I found that the naming and application of color prop defs was unintuitive and took me a bit to wrap my head around after we haven't worked in this area for a bit.

This refactors color prop defs so that:
- `inheritedColorPropDef` is renamed to `colorPropDef`, which should be our default colour prop def choice. This one doesn't yield any `data-accent-color` attribute when there is no explicit value.
- The original `colorPropDef` renamed to `accentColorPropDef`. This one yields an empty `data-accent-color` attribute when there is no explicit value, which is used for coloured text-based components and when we need to indicate that nested high-contrast text should use `--accent-12`

Some components also had an unnecessary `data-accent-color` attribute, which this PR removes

### DIFF
--- a/packages/radix-ui-themes/src/components/avatar.props.ts
+++ b/packages/radix-ui-themes/src/components/avatar.props.ts
@@ -1,6 +1,6 @@
 import {
   asChildPropDef,
-  colorPropDef,
+  accentColorPropDef,
   highContrastPropDef,
   radiusPropDef,
 } from '../props/index.js';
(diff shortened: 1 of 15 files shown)


benoitgrelard (approved):
That answers one of my question on the previous PR 😀

Yeah I think it's clearer now.

PR #509, opened 2024-05-25
fix usage of `inert` for React 19

chrisirhc:
## Description

https://github.com/facebook/react/pull/24730 has landed to React 19. 
When using radix-ui/themes with React 19, and using the Skeleton component, I get a warning that looks like:

```
Warning: Received an empty string for a boolean attribute `inert`. This will treat the attribute as if it were false. Either pass `false` to silence this warning, or pass `true` if you used an empty string in earlier versions of React to indicate this attribute is true.
    at span
    at eval (webpack-internal:///(ssr)/./node_modules/@radix-ui/themes/dist/esm/components/skeleton.js:11:179)
```

## Testing steps

Ran my application with this fix and no longer saw the warning.

## Relates issues / PRs

https://github.com/facebook/react/pull/24730


vladmoroz:
Thanks for the contribution – took a note of your implementation. We'll be looking at supporting React 19 holistically in #526

PR #526, opened 2024-06-11
React 19 support (RC)

vladmoroz:
- Pin deps (`@types/react` was having a hiccup with the current lockfile, removed the lockfile to resolve that, which triggered a bunch of other deps-related changes, like `turbo` upgrade. I resolved the latter but pinned all deps to specific versions to avoid this in the future)
- Upgrade Primitives to the new versions that support React 19 and RSC out of the box
- Fix `inert` in React 19
- Add React 19 to peer deps


packages/eslint-config-custom/package.json
```diff
@@ -4,10 +4,10 @@
   "main": "index.js",
   "license": "MIT",
   "dependencies": {
-    "eslint-config-next": "^13.4.1",
-    "eslint-config-prettier": "^8.3.0",
+    "eslint-config-next": "14.1.0",
+    "eslint-config-prettier": "8.3.0",
     "eslint-plugin-react": "7.28.0",
-    "eslint-config-turbo": "^1.9.3"
+    "eslint-plugin-turbo": "2.0.3"
```
lucasmotta, line 10: Is this package change intentional? If so we should also update the `eslint-config-custom` to use it as a plugin instead of a config.

lucasmotta (approved):
One small thing in regards the eslint config, other than that looks good.

packages/eslint-config-custom/package.json
vladmoroz: I had an issue when it requested I'd use the plugin, but rolling it back and just upgrading eslint-config-turbo it's gone now

Probably was a stale lockfile thing

PR #537, opened 2024-07-01
add support for named grid areas

fauxparse:
## Description

I use named grid areas all the time to simplify complex responsive layouts. I thought it'd be nice to be able to define these simply.

* Adds a `template` prop to the `Grid` component to set the `grid-template` CSS property
* Adds a `gridArea` layout prop

## Testing steps

Example usage in `apps/playground/app/test-grid-area/page.tsx`


packages/radix-ui-themes/src/components/grid.css
```diff
@@ -7,4 +7,10 @@
   justify-content: flex-start;
   grid-template-columns: minmax(0, 1fr);
   grid-template-rows: none;
+
+  &:where(.rt-r-gt) {
+    grid-template-columns: unset;
+    grid-template-rows: unset;
+    grid-template: var(--grid-template);
+  }
```
vladmoroz, line 15: - This should be a separate utility in `grid-template.css` wrapped in `@breakpoints`
- Are `grid-template-columns` and `grid-template-rows` needed? iirc `grid-template` shorthand would overwrite the longhand values if they are present, wouldn't it?

packages/radix-ui-themes/src/props/layout.props.ts
```diff
@@ -342,6 +342,22 @@ const layoutPropDefs = {
...
+   *
+   * @link
+   * https://developer.mozilla.org/en-US/docs/Web/CSS/grid-area
+   */
+  gridArea: {
+    type: 'string',
+    className: 'rt-r-ga',
+    customProperties: ['--grid-area'],
+    responsive: true,
+  },
```
vladmoroz, line 360: Let's put this prop in the alphabetical order with other grid props

packages/radix-ui-themes/src/components/grid.css
fauxparse: Good thought, fixed.

packages/radix-ui-themes/src/props/layout.props.ts
fauxparse: Fixed

fauxparse:
@vladmoroz Any further comments?

fauxparse:
@vladmoroz Any more to do here?

vladmoroz:
@fauxparse I took a closer look and adjusted a few things so that it works inline with the rest of the layout API

Basically, instead of the larger `grid-template` shorthand and `<Grid template="...">` prop, I changed to `grid-template-areas` and `<Grid areas="...">`. This feels more complementary alongside the existing `grid-template-columns` and `grid-template-rows`; also defining the track lengths in `grid-template` is just too much in JSX.

I'll merge now and release—thanks for the contribution

vladmoroz (approved):

PR #541, opened 2024-07-07
(Draft) Add intersection observer to skeleton

tresabhi:
## Description

Add intersection observer API directly into Skeleton. When `onIntersection` is passed, an observer is created. Passing `intersectionOptions` is also supported.

## Testing steps

Figuring this out as we speak.

## Relates issues / PRs

N/A

### DIFF
--- a/packages/radix-ui-themes/src/components/segmented-control.tsx
+++ b/packages/radix-ui-themes/src/components/segmented-control.tsx
@@ -1,16 +1,16 @@
 'use client';
 
-import * as React from 'react';
-import classNames from 'classnames';
 import * as ToggleGroupPrimitive from '@radix-ui/react-toggle-group';
 import { useControllableState } from '@radix-ui/react-use-controllable-state';
+import classNames from 'classnames';
+import * as React from 'react';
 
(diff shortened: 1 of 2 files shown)


vladmoroz:
Why is this PR created? What are you trying to achieve?

tresabhi:
@vladmoroz I am working on implementing the intersection observer right into the skeleton component. This would be very useful in many of my chases where I lazy load components as soon as the skeleton is visible on screen on scroll. So, instead of setting up an intersection observer wrapper, I was thinking of adding an `onIntersection` event directly to skeleton.

vladmoroz:
@tresabhi what does your design look like when things are loading? Do you granularly fetch data at the node level?

So far I'm not sure whether this makes sense in the Skeleton itself:
- If it's a large chunk of data that renders something like a big list or table, then the responsibilities wouldn't fall on an individual Skeleton element. The data loading and viewport intersection responsibilities would probably make more sense at the parent level
- If it's a single element off screen, it's either not worth the optimisation to lazy load a small piece of data, or is a pretty niche case

PR #557, opened 2024-08-05
fix: Tooltip: render non-string content outside of the Text component

evanfrawley:
## Description

When using the `<Tooltip>` component and passing in a non-string `React.ReactNode` into `content`, React produces warnings noting that specific elements cannot appear as descendants of a `p` tag. This change introduces a prop on `Tooltip` that allows the dev to define what `<Text as="_">` property they want to pass, enabling non-string content to be an allowed child in the tooltip and not produce a console warning.

The change introduces a `textAs` prop which pulls from the `textPropTypes['as']` value and passes that down to the `<Text>` in the tooltip, falling back to the currently hardcoded `p` value

Here are the warnings that are produced without this change:

CleanShot 2024-08-05 at 10 44 31@2x

CleanShot 2024-08-05 at 10 44 47@2x

## Testing steps

Manual testing on the playground page with a non-string ReactNode example

### DIFF
--- a/apps/playground/app/sink/page.tsx
+++ b/apps/playground/app/sink/page.tsx
@@ -237,6 +237,23 @@ export default function Sink() {
                                 Multiline
                               </Button>
                             </Tooltip>
+
+                            <Tooltip content={
+                               <DataList.Root my="1">
+                                <DataList.Item>
+                                  <DataList.Label style={{ color: 'var(--gray-5)'}}>Name</DataList.Label>
+                                  <DataList.Value style={{ color: 'var(--gray-1)'}}>Susan Kare</DataList.Value>
+                                </DataList.Item>
+                                <DataList.Item>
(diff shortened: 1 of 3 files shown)


evanfrawley:
I'm not quite sure if this is the correct solution -- I think another viable solution would be to have another prop, something like `contentAs` where that can be one of `span div label p` as found in `text.tsx`

chaance:
This is actually intentional. `Tooltip` is constrained by design to support inline (phrasing) content. Adding support for block content opens up a can of worms for accessibility. I don't think we want to deal with that in Themes, since it's opinionated by design.

If you are sure you need nested block content (most of the time you probably don't, which is why we have HoverCard), you can build your own using Radix Primitives. I'd advise you to vet this carefully and test extensively, expecially if you are using interactive content.

PR #566, opened 2024-08-15
[`TextArea`]: Accept number values for `Textarea`

### LINKED ISSUE #564: [TextArea] Issue
`TextArea` props don't match native the `textarea` props. 

Radix Themes: 
- value accepts `string | undefined`
- defaultValue accepts `string | undefined`

Native: 
- value accepts `string | number | readonly string[] | undefined`
- defaultValue `string | number | readonly string[] | undefined`

kevinmitch14:
## Description

Accept number to match native textarea more closely. 

## Relates issues / PRs
Closes #564

### DIFF
--- a/packages/radix-ui-themes/src/components/text-area.tsx
+++ b/packages/radix-ui-themes/src/components/text-area.tsx
@@ -11,8 +11,8 @@ import type { GetPropDefTypes } from '../props/prop-def.js';
 
 type TextAreaElement = React.ElementRef<'textarea'>;
 type TextAreaOwnProps = GetPropDefTypes<typeof textAreaPropDefs> & {
-  defaultValue?: string;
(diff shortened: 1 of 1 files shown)


kevinmitch14:
This is basically a mirror of #369

chaance:
Practically speaking I'm not sure you would ever want a number value for a textarea, which I think is probably the reason we kept this as-is. I'm inclined in this case to say it's better and more consistent with user expectations, even if it deviates from React's types for the native textarea. I could be convinced if you have a use-case for it though.

kevinmitch14:
I think you're right, it's not a quality of life improvement or anything, just keeping TS happy. No harm wrapping a `String()` around the value. Use case is just populating the TextArea with a value from DB that could be `string | number`

Issue #591, opened 2024-09-22
[Container] Problem with CSS prioritization


mickey8121:

mickey8121:
Hi there!
Please explain to me what is the point that the larger size is always prioritized for nested containers? Although, it just won't work since it's constrained by the parent container (if a container with `size={4}` is nested in a container with `size={3}`, the former can't be bigger than the latter anyway)

<img width="550" alt="Screenshot 2024-09-22 at 2 30 37 PM" src="https://github.com/user-attachments/assets/05d7223f-265b-46ca-9a21-b72d60b9c6c4">
<img width="548" alt="Screenshot 2024-09-22 at 2 29 52 PM" src="https://github.com/user-attachments/assets/d57072ab-0966-42d8-a41c-34abe34cdabc">

PR #593, opened 2024-09-23
fix: allow nesting sized containers

joeholdcroft:
## Description

Currently, if you try to nest a `Container` component with a smaller `size` than a `Container` component that wraps it at any level, the `size` does not correctly take effect. It always uses the largest `size` of any wrapping `Container` component due to CSS specificity.

This PR resolves this by applying the CSS for the sizing specifically to `.rt-ContainerInner` that are a **direct descendent** of the `.rt-Container`.

## Testing steps

Use the following JSX and check the correct widths are applied in browser.

```jsx
<Container size="4">
  <Text>This should be size 4</Text>
  <Container size="1">
    <Text>This should be size 1</Text>
  </Container>
</Container>
```

Alternatively, review the new Container samples in the "sink" demo example page.

## Related issues / PRs

N/A

### DIFF
--- a/apps/playground/app/sink/page.tsx
+++ b/apps/playground/app/sink/page.tsx
@@ -5893,6 +5893,37 @@ export default function Sink() {
                       </Flex>
                     </Flex>
                   </DocsSection>
+
+                  <DocsSection title="Container">
+                    <Text as="p" my="5">
+                      <Code>size</Code> can be set on nested <Code>Container</Code> instances:
+                    </Text>
+
+                    <Container size="4">
+                      <Box
+                        style={{
+                          backgroundColor: 'var(--color-panel-solid)',
+                          borderRadius: 'var(--radius-2)',
+                          boxShadow: 'inset 0 0 0 1px var(--gray-a4)',
+                        }}
+                        p="2"
+                      >
+                        <Text>This should be size 4</Text>
+                      </Box>
+                      <Container size="1">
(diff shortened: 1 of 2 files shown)


chaance:
Hmm, not entirely sure about the solution here. Having additional wrapping components will produce the same problematic results, no?

```tsx
<Container size="4">
  <Text>This should be size 4</Text>
  <Box>
    <Container size="1">
      <Text>This should be size 1</Text>
    </Container>
  </Box>
</Container>
```

I'll play around with this a bit and see what we can do.

chaance:
Ah yeah I think I missed that. I'll test locally and follow up as needed 👍

keithburgie:
@chaance, I came across two open issues about this (#606 and #591) and was excited to submit exactly the same fix. @joeholdcroft's PR correctly solves the problem!

joeholdcroft:
@chaance anything I can do to help get this one through? 🙏

chaance:
Still on my radar, just backed up with other priorities and need to spend a bit of time testing new features. Long backlog, thanks for your patience!

stevesims:
I just came across this exact same bug, and worked out the exact same fix

As the nature of a Container is that it will place a ContainerInner as its direct descendant, the solution in this PR is clearly the correct one to the underlying problem.  it's kinda disappointing that this fix hasn't been merged in and released yet

PR #600, opened 2024-10-03
[Popover]: remove min-width on `Popover.Content`

kevinmitch14:
## Description
Removed min-width CSS on `Popover.Content`. This was causing 2 issues as far as I can tell.
1. When the Popover.Trigger is wide, the content of the Popover will stretch to the same width of the trigger (`--radix-popover-trigger-width`)
2. Even when explicitly stating `<Popover.Content maxWidth="120px"/>`, the Content will stretch to the same size as the `Trigger`.

Screenshot 2024-10-03 at 12 26 36

### DIFF
--- a/packages/radix-ui-themes/src/components/popover.css
+++ b/packages/radix-ui-themes/src/components/popover.css
@@ -4,7 +4,6 @@
   outline: 0;
   overflow: auto;
   position: relative;
-
   --inset-padding-top: var(--popover-content-padding);
   --inset-padding-right: var(--popover-content-padding);
   --inset-padding-bottom: var(--popover-content-padding);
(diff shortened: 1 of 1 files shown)


chaance:
I think having a min-width of the trigger size is a good default for most cases. I think it might be better to have a separate custom property for the min-width set to the trigger width by default, but users can more easily override it directly this way. 

```diff
-  min-width: var(--radix-popover-trigger-width);
+  --radix-popover-min-width: var(--radix-popover-trigger-width);
+  min-width: var(--radix-popover-min-width);
```

Thoughts?

kevinmitch14:
Hmm, this still runs into the issue where if you define `maxWidth={"200px"}` and the trigger is 300px wide, the `Popover.Content` will be 300px 🤔

chaance:
I actually still think this is the best default, and I don't think there's any reason to add a new custom property. If you want to override this you can do so with a wrapper component for the Popover that sets the minWidth directly.

```tsx
import * as Popover from '@radix-ui/themes';

function PopoverContent(props) {
  return <Popover.Content minWidth="initial" {...props} />;
}
```

PR #602, opened 2024-10-03
Update dependencies and use semver range

chaance:
We recently released updated Radix primitives, so I'm updating all of our dependencies in Radix Themes to use the latest versions.

I also think it's better to use semver ranges rather than pin to specific versions. It's common for Themes users to also use some primitives directly, and if they pin their primitives to another version they'll end up with multiple copies of compatible packages. We do exactly this at WorkOS and it's a bit of a footgun if we don't update in the right order.

Since we follow semver patterns in our primitives this should be safe, and users can always choose to pin their dependencies if they run into issues.


packages/radix-ui-themes/package.json
```diff
@@ -84,41 +84,41 @@
...
+    "@radix-ui/react-aspect-ratio": "^1.1.0",
+    "@radix-ui/react-avatar": "^1.1.1",
+    "@radix-ui/react-checkbox": "^1.1.2",
+    "@radix-ui/react-compose-refs": "^1.1.0",
+    "@radix-ui/react-context": "^1.1.1",
+    "@radix-ui/react-context-menu": "^2.2.2",
+    "@radix-ui/react-dialog": "^1.1.2",
+    "@radix-ui/react-direction": "^1.1.0",
+    "@radix-ui/react-dropdown-menu": "^2.1.2",
     "@radix-ui/react-form": "0.1.0",
```
chaance, line 100: Left this one pinned since it's not a stable release. We may be able to go ahead and ship a 1.0 since I haven't seen any major issues with the API, and we haven't planned anything around this AFAIK.

lucasmotta: I didn't even know this was a dependency 🧐

chaance: Hah, we aren't even using it. I'll just remove!

lucasmotta (approved):

PR #603, opened 2024-10-03
Add `minimal` variant for `Table`

chaance:
The new `minimal` table variant includes no borders between cells, and first/last cells will have their left/right padding removed respectively.

Screenshot of the new variant adjacent to the other variants at each size:

CleanShot 2024-10-03 at 14 10 40@2x

### DIFF
--- a/apps/playground/app/sink/page.tsx
+++ b/apps/playground/app/sink/page.tsx
@@ -232,7 +232,7 @@ export default function Sink() {
                               </Button>
                             </Tooltip>
 
-                            <Tooltip content="The goal of typography is to relate font size, line height, and line width in a proportional way that maximizes beauty and makes reading easier and more pleasant.">
+                            <Tooltip content="The goal of typography is to relate font size, line height, and line width in a proportional way that maximizes beauty and makes reading easier and more pleasant.">
                               <Button variant="solid" size="1">
                                 Multiline
                               </Button>
@@ -3770,7 +3770,7 @@ export default function Sink() {
                                   Typography
                                 </Text>
                                 <Text as="p" color="gray" size={size} style={{ maxWidth: '40ch' }}>
-                                  The goal of typography is to relate font size, line height, and
+                                  The goal of typography is to relate font size, line height, and
                                   line width in a proportional way that maximizes beauty and makes
                                   reading easier and more pleasant.
                                 </Text>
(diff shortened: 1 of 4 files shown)


apps/playground/app/sink/page.tsx
```diff
@@ -232,7 +232,7 @@ export default function Sink() {
                               </Button>
                             </Tooltip>
 
-                            <Tooltip content="The goal of typography is to relate font size, line height, and line width in a proportional way that maximizes beauty and makes reading easier and more pleasant.">
+                            <Tooltip content="The goal of typography is to relate font size, line height, and line width in a proportional way that maximizes beauty and makes reading easier and more pleasant.">
```
chaance, line 235: There were a bunch of weird, inconsistent whitespace characters in this file so I just replaced them all.

packages/radix-ui-themes/src/components/table.css
```diff
@@ -32,16 +40,17 @@
...
-  padding: var(--table-cell-padding);
+  padding: var(--table-cell-padding-top) var(--table-cell-padding-right) var(--table-cell-padding-bottom)
+    var(--table-cell-padding-left);
   /* Works as min-height */
   height: var(--table-cell-min-height);
 
   /* Inset with Table */
   .rt-Inset :where(&:first-child) {
-    padding-left: var(--inset-padding-left, var(--table-cell-padding));
+    padding-left: var(--inset-padding-left, var(--table-cell-padding-left));
```
chaance, line 50: This fixes a potential (but unlikely) bug, where the fallback value would be invalid in cases where `--table-cell-padding` evaluates to a shorthand value.

packages/radix-ui-themes/src/components/table.css
```diff
@@ -132,3 +153,37 @@
   --scrollarea-scrollbar-horizontal-margin-left: 0;
   --scrollarea-scrollbar-horizontal-margin-right: 0;
 }
+
+/* minimal */
+
+.rt-TableRoot:where(.rt-variant-minimal) {
```
lucasmotta, line 159: I thinking if we can simplify the changes here to achieve the `minimal` visuals, while keeping backwards compatible:

1. Move some of the variables to `.rt-TableRoot` and create a new variable:
```scss
.rt-TableRoot {
  --table-cell-edge-padding: var(--table-cell-padding-inline);
  --table-row-box-shadow: inset 0 -1px var(--gray-a5);
  --table-row-background-color: transparent;
}
```

2. Update the `.rt-TableCell` to apply a specific padding to the left/right side of the first/last child and to use `padding-block/padding-inline` (this is because size-3 uses different padding for block/inline):

```scss
&:where(&:first-child) {
  padding-left: var(--table-cell-edge-padding, var(--table-cell-padding-inline));
}

&:where(&:last-child) {
  padding-left: var(--table-cell-edge-padding, var(--table-cell-padding-inline));
}

/* Inset with Table */
.rt-Inset :where(&:first-child) {
  padding-left: var(--inset-padding-left, var(--table-cell-padding-inline));
}
.rt-Inset :where(&:last-child) {
  padding-right: var(--inset-padding-right, var(--table-cell-padding-inline));
}
```

3. Update the sizes to use the new variables:
```scss
// size 1
--table-cell-padding-inline: var(--space-2);

// size 2
--table-cell-padding-inline: var(--space-3);

// size 3
--table-cell-padding-inline: var(--space-4);
--table-cell-padding-block: var(--space-3);
```

3. Create the minimal variant:

```scss
.rt-TableRoot:where(.rt-variant-minimal) {
  --table-row-box-shadow: none;
  --table-cell-edge-padding: 0px;

  --scrollarea-scrollbar-horizontal-margin-left: 0;
  --scrollarea-scrollbar-horizontal-margin-right: 0;
}
```

PR #621, opened 2024-10-29
Segmented control: Support `disabled` prop

kendallstrautman:
## Description

Extend `SegmentedControl` to support `disabled` functionality and style. In this implementation both `classic` and `surface` variants share the same disabled styles. [See design spec ](https://www.figma.com/design/Bdv2UWN3GrAH4qUIL7Ldnl/Launch-Week%3A-Radar?node-id=2687-22849&node-type=frame&t=S9FfI6CBMSrkXWkN-0)and [context for request](https://work-os.slack.com/archives/C066T3GNG4Q/p1729705862819629?thread_ts=1729704870.866939&cid=C066T3GNG4Q).

Screenshot 2024-10-29 at 2 08 23 PM
Screenshot 2024-10-29 at 2 08 28 PM

## Testing steps

Visit the [sink page in the playground](https://themes-playground-git-ksadd-segmented-control-disabled-workos.vercel.app/sink#segmented-control) to see the disabled control.

### DIFF
--- a/packages/radix-ui-themes/src/components/segmented-control.css
+++ b/packages/radix-ui-themes/src/components/segmented-control.css
@@ -20,6 +20,12 @@
 
   /* Create a new stacking context */
   isolation: isolate;
+
+  &.disabled {
+    cursor: var(--cursor-disabled);
+    color: var(--gray-a8);
+    background-color: var(--gray-3);
+  }
 }
 
 .rt-SegmentedControlItem {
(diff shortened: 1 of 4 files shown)


packages/radix-ui-themes/src/components/segmented-control.css
```diff
@@ -20,6 +20,12 @@
 
   /* Create a new stacking context */
   isolation: isolate;
+
+  &.disabled {
```
chaance, line 24: Should we use `:where` or is the added specificity useful here?
```suggestion
  &:where(.disabled) {
```

kendallstrautman: The specificity is needed here to override the `color` and `background-color` on the base class.

chaance (approved):

PR #624, opened 2024-11-07
Fix CSS specificity issues with `SegmentedControl`

chaance:
This fixes two existing issues with style linting:

```
src/components/segmented-control.css
  24:3  ✖  Expected "&.disabled" to have a specificity no more than "0,1,1"                                             selector-max-specificity
  24:4  ✖  Expected ".disabled" to match pattern "/^((xs|sm|md|lg|xl):)?-?rt-|^radix-themes$|^(light|dark)(-theme)?$/"  selector-class-pattern
```

### DIFF
--- a/packages/radix-ui-themes/src/components/segmented-control.css
+++ b/packages/radix-ui-themes/src/components/segmented-control.css
@@ -20,12 +20,11 @@
 
   /* Create a new stacking context */
   isolation: isolate;
+}
 
-  &.disabled {
-    cursor: var(--cursor-disabled);
-    color: var(--gray-a8);
-    background-color: var(--gray-3);
-  }
+.rt-SegmentedControlRoot:where([data-disabled]) {
+  color: var(--gray-a8);
+  background-color: var(--gray-3);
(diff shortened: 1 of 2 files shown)


packages/radix-ui-themes/src/components/segmented-control.css
```diff
@@ -20,12 +20,11 @@
...
   /* Create a new stacking context */
   isolation: isolate;
+}
 
-  &.disabled {
-    cursor: var(--cursor-disabled);
-    color: var(--gray-a8);
-    background-color: var(--gray-3);
-  }
+.rt-SegmentedControlRoot:where([data-disabled]) {
```
chaance, line 25: Not idea why, but when nesting the selector, one of our postcss plugins is moving this declaration before the parent declaration. Un-nesting fixes this, so we don't need extra specificity to override this.

Also: using `data-disabled` handles `cursor` for us since that's set everywhere by our base styles.

kendallstrautman (approved):
ty!

PR #647, opened 2024-12-17
Expose components from standalone entrypoints

### LINKED ISSUE #645: [Next.js] Entire `@radix-ui/themes` is being bundled into the client bundle when compound components are being used
As soon as any compound components, e.g. `Popover.*` is used, the entire `@radix-ui/themes` will be bundled into the client bundle of the web app, even when [optimizePackageImports](https://nextjs.org/docs/app/api-reference/config/next-config-js/optimizePackageImports) is specified.

**Minimal reproduction:** https://github.com/mwskwong/next-radix-theme-compound-components-bundling

**Step to reproduce:**
1. Run `npm run build` to trigger a PROD build, a bundle analyze report will be auto-generated
2. From the client bundle report `/.next/analyze/client.html`, observe that the entire `@radix-ui/theme` is being bundled, even though only a few components are used.
3. Navigate to `src/app/page.tsx`
4. Comment/remove the usage of `Popover`
5. Repeat steps 1 - 2 again
6. Observe that only necessary components are being bundled this time.

**Context:**
This seems to be a common issue related to how bundling works in Next.js. Some other component libraries like Mantine [offer separate imports to import those components as regular components](https://help.mantine.dev/q/server-components#error-compound-components-in-server-component), while it doesn't seem to be possible in Radix Themes, at least this is how the TS type suggests

**Workaround**
Directly import from `popover.js` instead.
```js
import * as Popover from '@radix-ui/themes/dist/esm/components/popover.js';
```

Obviously, this is not how it is designed to work, and things may randomly break in the future. But for a simple case like this, it seems to be working well.

chaance:
This PR adds package exports for all public components and utilities. Doesn't automatically fix #645 as this is a bundler problem, but it gives consumers a way to optimize their own bundles when treeshaking fails. Internal modules in the components directory have been moved into an `_internal` folder and are not considered public API.

```tsx
// Components previously imported from from '@radix-ui/themes'
import { AlertDialog } from '@radix-ui/themes/alert-dialog';
import { Box } from '@radix-ui/themes/box';
// We recommend using namespaced imports for compound components
import * as CheckboxGroup from '@radix-ui/themes/checkbox-group';
// Utilities previously exported from '@radix-ui/themes/props'
import { dialogContentPropDefs } from '@radix-ui/themes/dialog.props.js';
```

## Relates issues / PRs

#645


package.json
```diff
@@ -6,7 +6,7 @@
     "build": "turbo run build",
     "dev": "turbo run dev --no-cache --continue",
     "lint": "turbo run lint",
-    "clean": "turbo run clean && rm -rf node_modules"
+    "clean": "turbo run clean && rm -rf node_modules && rm -rf .turbo"
```
chaance, line 9: Ran into an issue with rc v2 where turbo cache skipped the build after clearing build artifacts, so this fixes that. Probably gonna nuke turborepo in a follow, no reason to even need it here tbh.

hadihallak (approved):

PR #667, opened 2025-01-03
add support for custom select indicator

fauxparse:
## Description

Adds the ability to provide a custom indicator to the `Select` component, replacing the default checkmark. This is useful in situations where you're using a custom icon set.

## Testing steps

Provide an `indicator` prop to `Select.Root`. This should be a `ReactNode`. An example is given in `sink.tsx`.


chaance:
No plans to add this level of customization at this point. I recommend building a custom version of the select using the underlying Radix Primitives if you want to deviate from the options currently provided.

PR #734, opened 2025-06-10
feat: theme high contrast

ihardlight:
## Description

To improve accessibility and better support users with visual impairments, this PR introduces first-class support for high contrast themes. According to accessibility guidelines, high contrast modes are essential for many users, and making them easier to enable and manage significantly improves usability and inclusiveness.

## Changes

- Added a highContrast prop to the Theme components. When enabled, all nested elements automatically receive high contrast styling, removing the need to configure this individually on each component.
- Added a high contrast toggle to ThemeToggle, making it easy to preview and switch high contrast mode during development or runtime.
- When high contrast mode is active, the attribute `[data-high-theme="true"]` and the `.high-contrast` class are applied. This allows developers to check the contrast state in CSS or toggle it using external libraries—similar to the existing `.light` and `.dark` class mechanisms.

## Testing steps

Open playground and try to toggle high contrast or copy theme
Critical moment: if you enable global high contrast you can not turn it off for individual component

## Relates issues / PRs

Add the same functionality to next-themes https://github.com/pacocoursey/next-themes/pull/360 to toggle contrast value using `.high-contrast` class in SSR-friendly way and including support `prefers-contrast:more` media query

In future step, after this PR will be merged I will update radix-ui docs to add info about this functional
After next-theme PR will be merged and released I will add an example how it works in playground

### DIFF
--- a/packages/radix-ui-themes/.stylelintrc.cjs
+++ b/packages/radix-ui-themes/.stylelintrc.cjs
@@ -6,7 +6,7 @@ module.exports = {
     // This is so that Tailwind classes work as expected.
     'selector-max-specificity': ['0,1,1'],
     // Enforce prefixes on classnames and keyframes
-    'selector-class-pattern': /^((xs|sm|md|lg|xl):)?-?rt-|^radix-themes$|^(light|dark)(-theme)?$/,
+    'selector-class-pattern': /^((xs|sm|md|lg|xl):)?-?rt-|^radix-themes$|^(light|dark)(-theme)?$|^more-contrast$/,
     'keyframes-name-pattern': /^rt-([a-z]|-)+$/,
   },
 };
--- a/packages/radix-ui-themes/.stylelintrc.cjs
+++ b/packages/radix-ui-themes/.stylelintrc.cjs
@@ -6,7 +6,7 @@ module.exports = {
     // This is so that Tailwind classes work as expected.
     'selector-max-specificity': ['0,1,1'],
     // Enforce prefixes on classnames and keyframes
-    'selector-class-pattern': /^((xs|sm|md|lg|xl):)?-?rt-|^radix-themes$|^(light|dark)(-theme)?$|^more-contrast$/,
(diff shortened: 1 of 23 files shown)


packages/radix-ui-themes/.stylelintrc.cjs
```diff
@@ -6,7 +6,7 @@ module.exports = {
     // This is so that Tailwind classes work as expected.
     'selector-max-specificity': ['0,1,1'],
     // Enforce prefixes on classnames and keyframes
-    'selector-class-pattern': /^((xs|sm|md|lg|xl):)?-?rt-|^radix-themes$|^(light|dark)(-theme)?$/,
+    'selector-class-pattern': /^((xs|sm|md|lg|xl):)?-?rt-|^radix-themes$|^(light|dark)(-theme)?$|^more-contrast$/,
```
ihardlight, line 9: should work if body has `.more-contrast ` class

lucasmotta (commented):
I like this idea, but can we keep only the `:where([data-high-contrast="true"])` selector?
I see the `.more-contrast` may be provided by next-themes, but since that's not merged yet we can add this back later if needed.

ihardlight:
@lucasmotta I believe it’s a good idea to establish a consistent API for themes and contrasts from the beginning. Both can be accessed by modifying the class of the body tag.

Additionally, I’ve separated the next-contrast and next-themes. They’re still in the same repository but under different npm packages, allowing me to release them promptly without merging a PR.

In my opinion, it’s more practical to maintain the class toggle for contrast.

ihardlight:
@lucasmotta I have resolved your issue. Would you like to proceed with merging PR?

chaance:
I'd like to add a global high-contrast feature as well but I'm not sure at this time it's as simple as this. There are gaps in our current APIs for high-contrast that I don't think satisfy a11y needs. I'm going to close this for now but we'll be revisiting the idea soon.

PR #797, opened 2026-02-12
DRAFT: Add support for container queries

chaance:


.cursor/plans/container_query_support_b639a7ba.plan.md
```diff
@@ -0,0 +1,379 @@
...
+**[extract-props.ts](packages/radix-ui-themes/src/helpers/extract-props.ts)**: The existing extraction logic calls `getResponsiveClassNames` / `getResponsiveStyles` which will handle `@container` automatically once updated. The code that applies prop def defaults only reads `value.initial`, so the `@container` sub-object passes through untouched.
+
+### 3. PostCSS plugin — [postcss-breakpoints.cjs](packages/radix-ui-themes/postcss-breakpoints.cjs)
+
+Extend the existing `@breakpoints` directive so that for each rule, it generates **both** media query and container query variants:
+
+- **Media (existing)**: `.sm\:rt-r-display-flex` inside `@media (min-width: 768px)`
+- **Container (new)**: `.cq-sm\:rt-r-display-flex` inside `@container (min-width: 768px)`
+
+**Container query at-rules are placed after all media query at-rules** to ensure they take precedence in the cascade when both apply.
```
chaance, line 162: I'm not 100% sure this is always what a user would expect, but I'm also not sure you could ever know for sure in all cases what a user would expect. Seems we have to make a decision one way or another.

